PR #3, opened 2024-04-16
Add basic Gradle Library Implementation

linglejack06:
**Summary**: Create a basic gradle library to hold common code across years starting with utils folder

**Changes**:
- add WPILIb implementation dependencies
- add WPILib plugins
- copy utils folder from 2024-robot-code


linglejack06:
@jkleiber I've got most things migrated over, however, I am having some difficulty with the build.gradle.kts file, do you have experience creating/modifying the file?

jkleiber:
@linglejack06 I've modified build.gradle a lot before, but I'm not sure what build.gradle.kts is. I'll look into it some, but lmk if you figure it out

jkleiber:
@linglejack06 check this link out: https://docs.github.com/en/actions/publishing-packages/publishing-java-packages-with-gradle

linglejack06:
@jkleiber ill check the link out. The .kts is just kotlin script version so it has slightly different syntax ( I can figure that part out ) I just needed some help on the implementation vs dependency etc

jkleiber:
@linglejack06 I think we should just keep the problematic wpilib deps out for now and figure it out later

linglejack06:
@jkleiber okay I can do that, I just wont be able to test anything because several util classes utilize the pose 2d class or units class. Ill  remove them, and I will see if I can get them working. I assume its just an error on how I am loading them in.

jkleiber:
@linglejack06 anything using Pose we can decompose as needed into coordinates/rotation. For units, we could depend on yet another units library or add some of our own conversion constants

linglejack06:
@jkleiber everything has a basic implementation now, build.gradle kinks were worked out and command library was added for WaitForButton and feedforward tune commands. This should be ready to be merged into main

lib/src/main/java/coppercore/commands/WaitForButtonCommand.java
```diff
@@ -0,0 +1,32 @@
+package coppercore.commands;
```
jkleiber, line 1: Do we actually use this? It seems like this would be redundant to wpilib's onTrue function

lib/src/main/java/coppercore/utils/FieldFinder.java
```diff
@@ -0,0 +1,169 @@
+package coppercore.utils;
```
jkleiber, line 1: This can go in as-is for now, but could you create a ticket to generalize this logic please

jkleiber (commented):
A couple comments before we merge

lib/src/main/java/coppercore/utils/FieldFinder.java
linglejack06: yes I was planning on updating that and a few others such as the Field Constants (as it basically all relates to this year specifically )

lib/src/main/java/coppercore/commands/WaitForButtonCommand.java
linglejack06: I agree by looking at it, however, I left it in since I figured it was used somewhere. Ill see if i can find any uses of it in 2024 code

linglejack06:
@jkleiber command is removed as I did not find any mentions of it in robot container

jkleiber (approved):
approved!

PR #14, opened 2024-08-27
Deadband

linglejack06:
- add Deadband file in deadband-help package


linglejack06:
@jkleiber this should cause no issues, i just copy and pasted the code from 2024-robot-code into a file here since i recognized that it would be useful for all projects

jkleiber (commented):
@linglejack06 can you rebase this? It looks like it's running the old CI. Also can we put Deadband in the math library, and the other stuff in geometry

linglejack06:
Yeah I'll update that

jkleiber:
@linglejack06 once you resolve merge conflicts here I can take a look at reviewing

linglejack06:
@jkleiber merge conflicts are fixed

geometry/build.gradle
jkleiber, line 1: I don't think this should be deleted

jkleiber (commented):
Let's add back the geometry/build.gradle and then this can be merged

geometry/build.gradle
```diff
@@ -6,67 +6,36 @@
...
-	// Allow ignoring certain parts in formatting.
-	toggleOffOn()
-	// apply a specific flavor of google-java-format
-	googleJavaFormat('1.19.2').aosp().reflowLongStrings()
-	// fix formatting of type annotations
-	formatAnnotations()
-}
-}
-
-compileJava.dependsOn 'spotlessApply'
```
jkleiber, line 44: Can we keep the formatter pls

jkleiber (commented):
We still need to add the formatter back to this build.gradle

aidnem:
@linglejack06 can we fix spotless so that the CI will run?

linglejack06:
@aidnem yeah, I think I messed git up because I've tried adding spotless back five times on my local side and it works, but it's apparently not showing up here 😂

geometry/src/main/java/coppercore/FieldFinder.java
jkleiber, line 1: This should be in the other PR only

geometry/src/main/java/coppercore/Point.java
jkleiber, line 1: The geometry PR is leaking into this one somehow, maybe we can merge that first

geometry/src/main/java/coppercore/Triangle.java
jkleiber, line 1: geometry PR

jkleiber (commented):
We should remove the geometry PR files from this PR - `git restore <filename>` is probably your friend here

If we get the other PR merged and rebase this one, it will probably ignore those files since they would not have changed. But if anything changes then these files would reappear, so we should just fix it here

geometry/src/main/java/coppercore/Triangle.java
linglejack06: i think i have removed all geometry-related files other than build.gradle, not sure how that was all happening.

jkleiber (approved):

PR #17, opened 2024-09-03
hopefully working math and controls

minhnguyenbhs:


math/src/main/java/frc/robot/BuildConstants.java
```diff
@@ -0,0 +1,17 @@
+package frc.robot;
```
jkleiber, line 1: Can we git ignore and untrack this file please?

jkleiber (commented):
Only one comment

jkleiber (approved):

PR #18, opened 2024-09-04
Add json config

avidraccoon:
Adds ability to load constants from JSON file.


parameter_tools/build.gradle
jkleiber, line 1: Can we add the `spotless` formatter to this build.gradle so the code gets auto-formatted?

parameter_tools/src/main/java/coppercore/parameter_tools/JSONSync.java
```diff
@@ -0,0 +1,115 @@
...
+        this.file = file;
+    }
+
+    private static record JSONSyncConfig(boolean serializeNulls, boolean prettyPrinting, boolean excludeFieldsWithoutExposeAnnotation, FieldNamingPolicy namingPolicy,
+    ToNumberPolicy numberToNumberPolicy, ToNumberPolicy objectToNumberPolicy, LongSerializationPolicy longSerializationPolicy, boolean autoReload) {       
+        public JSONSyncConfig(JSONSyncConfigBuilder builder){
+            this(builder.serializeNulls, builder.prettyPrinting, builder.excludeFieldsWithoutExposeAnnotation, builder.namingPolicy, builder.numberToNumberPolicy,
+                builder.objectToNumberPolicy, builder.longSerializationPolicy, builder.autoReload);
+        }
+    }
```
jkleiber, line 55: Why is there a method within a method here?

jkleiber (commented):
First pass review

parameter_tools/src/main/java/coppercore/parameter_tools/JSONSync.java
avidraccoon: One is a record which is a data structure that is immutable and the other is a constructor for the record.

parameter_tools/build.gradle
avidraccoon: I believe we can, but I am not sure.

PR #20, opened 2024-09-16
Geometry

### LINKED ISSUE #4: Generalize Field Finder Util
**Summary**: Currently the field finder util is geared more specifically towards the 2024 field and related poses. It would be best to generalize this to be used across field designs from year-to-year

linglejack06:
- add fieldfinder utils
- including Is point in triangle
- and is point on triangle (either in or on surrounding lines)

resolves #4


geometry/build.gradle
```diff
@@ -41,9 +40,6 @@ java {
 }
 }
 
-compileJava.dependsOn 'spotlessApply'
```
jkleiber, line 44: can we add this back so formatting happens?

geometry/src/main/java/coppercore/FieldFinder.java
```diff
@@ -0,0 +1,36 @@
...
+        Triangle t1 = new Triangle(point, triangle.getPoint2(), triangle.getPoint3());
+        Triangle t2 = new Triangle(triangle.getPoint1(), point, triangle.getPoint3());
+        Triangle t3 = new Triangle(triangle.getPoint1(), triangle.getPoint2(), point);
+
+        double area = triangle.getArea();
+
+        double sumOfT = t1.getArea() + t2.getArea() + t3.getArea();
+
+        return Math.abs(sumOfT - area) < 0.05;
+    }
```
jkleiber, line 24: I feel like this function should be a part of a general `containsPoint` function that any `ConvexPolygon` (an abstract superclass name I just came up with) should be able to implement

geometry/src/main/java/coppercore/FieldFinder.java
```diff
@@ -0,0 +1,36 @@
...
+    public boolean isPointCollidingWithTriangle(Point point) {
+        Line line1 = new Line(triangle.getPoint1(), triangle.getPoint2());
+        Line line2 = new Line(triangle.getPoint2(), triangle.getPoint3());
+        Line line3 = new Line(triangle.getPoint3(), triangle.getPoint1());
+
+        return isPointInsideTriangle(point)
+                || line1.isPointOnLine(point)
+                || line2.isPointOnLine(point)
+                || line3.isPointOnLine(point);
+    }
```
jkleiber, line 35: See above comment on making point/convex polygon intersections a general function rather than specific to FieldFinder. This would be similar to what you've already done with lines

This will allow FieldFinder to extend beyond just triangles very easily

geometry/src/main/java/coppercore/Line.java
```diff
@@ -0,0 +1,31 @@
...
+    private double x2;
+    private double y2;
+    private double slope;
+
+    public Line(double x1, double y1, double x2, double y2) {
+        this.x1 = x1;
+        this.y1 = y1;
+        this.x2 = x2;
+        this.y2 = y2;
+        this.slope = (y2 - y1) / (x2 - x1);
```
jkleiber, line 15: What if x2==x1?

(The answer is probably to use linear interpolation and parameterize the line rather than using slope intercept form)

math/src/main/java/coppercore/Deadband.java
jkleiber, line 1: this file shouldn't be in this PR 👀

jkleiber (commented):
first pass review

geometry/build.gradle
linglejack06: its on line 68

geometry/src/main/java/coppercore/FieldFinder.java
linglejack06: added

geometry/src/main/java/coppercore/FieldFinder.java
linglejack06: done

math/src/main/java/coppercore/Deadband.java
linglejack06: oops

geometry/src/main/java/coppercore/Line.java
linglejack06: how would i go about using linear interpolation?

jkleiber: So I found a better answer actually:

For 3 points (A, B, and C) if you have distance(A,B) = distance (A, C) + distance(C, B), then C is on the line between A and B

So you just need to do the distance formula a few times, which honestly should be a function of a `Line` anyway.

geometry/src/main/java/coppercore/Line.java
```diff
@@ -0,0 +1,31 @@
...
+        this.y2 = point2.getY();
+        this.slope = (this.y2 - this.y1) / (this.x2 - this.x1);
+    }
+
+    public boolean isPointOnLine(Point point) {
+        return point.getY() == point.getX() * slope
+                && (x1 < point.getX() && point.getX() < x2)
+                && (y1 < point.getY() && point.getY() < y2);
+    }
+}
```
aidnem, line 31: Say we have a line that goes from (0,1) to (2,3)
The slope of that line would be 1.
However, with this function, if we test whether the point (0.5, 1.5) is on the line, the condition:
`point.getY() == point.getX() * slope` will be false because: 1.5 != 0.5 * 1, even though the point is on the line.
Could we reimplement this function to work differently? What @jkleiber wrote above should work.

geometry/src/main/java/coppercore/FieldFinder.java
```diff
@@ -0,0 +1,17 @@
...
+
+    public void setShape(ConvexPolygon polygon) {
+        this.polygon = polygon;
+    }
+
+    public boolean isRobotInShape(Point p) {
+        return this.polygon.containsPoint(p);
+    }
+}
```
aidnem, line 17: Do we want to make a FieldFinder be able to contain more than one polygon or do you think that would be unnecessary?

aidnem (changes requested):
Aside from the broken isPointOnLine function, this looks good :)
Once we fix that, request me to review and I'll look over it again and probably approve.

aidnem (approved):
Looks good to me

jkleiber:
Can we delete the slope variable before we merge this (it is unused and still is prone to divide by 0)

jkleiber (approved):

jkleiber:
@linglejack06 please merge this when able

PR #31, opened 2024-11-07
Add monitors

aidnem:
Adds:
- coppercore.monitors.Monitor
- coppercore.wpi_interface.MonitoredSubsystem
- unit tests for monitor


monitors/src/main/java/coppercore/monitors/Monitor.java
```diff
@@ -0,0 +1,205 @@
...
+
+        triggered = !isStateValid.getAsBoolean();
+        if (triggered) {
+            if (triggeredTime == 0.0) {
+                triggeredTime = currentTimeSeconds;
+            }
+
+            if (currentTimeSeconds - triggeredTime >= timeToFault) {
+                faulted = true;
+            }
```
jkleiber, line 68: If an if statement only sets a boolean value, then we should just make the boolean value get set directly, i.e.
```
faulted = (currentTimeSeconds - triggeredTime) >= timeToFault;
```

monitors/src/main/java/coppercore/monitors/Monitor.java
```diff
@@ -0,0 +1,205 @@
...
+        } else {
+            if (!sticky) {
+                faulted = false;
+            }
```
jkleiber, line 72: Same comment as above here: 
```
faulted = sticky;
```

monitors/src/main/java/coppercore/monitors/Monitor.java
```diff
@@ -0,0 +1,205 @@
...
+    public void periodic(double currentTimeSeconds) {
+        // currentTimeSeconds doesn't need to be from a specific point in history
+        // As long as the reference point is always the same, it could be from
+        // the robot being turned on, initialized, etc.
+
+        triggered = !isStateValid.getAsBoolean();
+        if (triggered) {
+            if (triggeredTime == 0.0) {
+                triggeredTime = currentTimeSeconds;
+            }
```
jkleiber, line 64: Coding practice: don't check floating precision numbers for equality

This is sort of a nit because it is comparing to 0.0. I would prefer if triggeredTime was a negative number by default and we compared with <= though.

monitors/src/test/java/MonitorTests.java
```diff
@@ -0,0 +1,91 @@
+package coppercore.monitors.test;
+
+import coppercore.monitors.Monitor;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class MonitorTests {
+    // TODO: FIGURE OUT HOW TO TEST THIS
+    // Callbacks are hard :(
```
jkleiber, line 9: could you make some variable owned by the test fixture and then increment it in the callback function? Or do something with a static variable?

wpi_interface/build.gradle
```diff
@@ -3,6 +3,7 @@ import java.text.SimpleDateFormat
 plugins {
 	id "java"
 	id "maven-publish"
+	id "edu.wpi.first.GradleRIO" version "2024.3.2"
```
jkleiber, line 6: not a fan of importing the whole GradleRIO if we can avoid it

wpi_interface/src/main/java/coppercore/wpi_interface/MonitoredSubsystem.java
```diff
@@ -0,0 +1,47 @@
...
+    public void periodic() {
+        monitoredPeriodic();
+        runMonitors();
+    }
+
+    /**
+     * OVERRIDE ME! This function is called every time the subsystem's periodic function is called.
+     * However, MonitoredSubsytem automatically checks monitors during every periodic run.
+     * Therefore, this method should be overridden as a replacement for the normal periodic function
+     * in the implementation of the subsystem.
```
jkleiber, line 27: Could we make MonitoredSubsystem an abstract class, and then force all users of MonitoredSubsystem to provide this implementation?

wpi_interface/src/main/java/coppercore/wpi_interface/MonitoredSubsystem.java
```diff
@@ -0,0 +1,47 @@
...
+
+    private void runMonitors() {
+        registeredMonitors.forEach(
+                monitor -> {
+                    monitor.periodic(Timer.getFPGATimestamp());
+
+                    Logger.recordOutput(
+                            "monitors/" + monitor.getName() + "/triggered", monitor.isTriggered());
+                    Logger.recordOutput(
+                            "monitors/" + monitor.getName() + "/faulted", monitor.isFaulted());
```
jkleiber, line 44: I think it would be good to be able to turn this logging on or off. We run out of memory with lots of strings and we will probably have a bunch of monitors this season.

Maybe the move is to make each monitor subsystem have a flag for if it should be logging these. I'm fine if this gets deferred to a different ticket, but it's just something to think about

jkleiber (commented):
first pass feedback

monitors/src/main/java/coppercore/monitors/Monitor.java
aidnem: What sentinel value would you prefer that I pick to indicate that it isn't currently triggered in place of zero?

wpi_interface/src/main/java/coppercore/wpi_interface/MonitoredSubsystem.java
aidnem: I've added the ability to enable/disable logging for the MonitoredSubsystem and to enable/disable logging for each individual monitor as well.

monitors/src/test/java/MonitorTests.java
aidnem: Done! Thanks for the suggestion on how to do this.

monitors/src/main/java/coppercore/monitors/Monitor.java
jkleiber: -1 is good

jkleiber (approved):
approved, unit tests are passing is my verification evidence

aidnem:
Sorry, I didn't see this approval or I would've merged it myself

PR #36, opened 2024-11-12
add drive command with smoother inputs

linglejack06:
- square linear velocity while retaining direction of x and y
- square omega input using Math.copySign to retain direction
- rename wpi_interface to wpilib_interface


wpi_interface/src/main/java/coppercore/wpi_interface/DriveWithJoysticks.java
```diff
@@ -0,0 +1,47 @@
...
+
+        Rotation2d direction = new Rotation2d(x, y);
+        double squaredMagnitude = deadband * deadband;
+
+        Translation2d linearVelocity =
+                new Pose2d(new Translation2d(), direction)
+                        .transformBy(new Transform2d(squaredMagnitude, 0.0, new Rotation2d()))
+                        .getTranslation();
+
+        return linearVelocity;
```
jkleiber, line 45: This logic is sus to me, and definitely needs to be tested

If it works, we will need some comments added to explain the logic and why it works

jkleiber (commented):
Just one comment. Any sim evidence would be good before we put this on the robot

jkleiber (approved):

PR #40, opened 2024-11-23
Basic vision subsystem

aidnem:
Adds the vision subsystem from high-key-2024 but with some improvements:
- Removes the trust zone system. This means that later we can implement actual calculation of standard deviations.
- Makes the field layout parametrized: The robot code can choose which tags to pass to the vision system.
- Use Pose3d instead of Pose2d whenever possible. This moves the step of converting the vision poses back to 2d out of the vision system and into the robot project wherever they're needed as Pose2d's, preserving the additionally data for as long as possible.
- Move camera params out of core vision constants and add it as a parameter so they can be defined individually in each robot project.


aidnem:
Here's verification evidence of it working in sim:

https://github.com/user-attachments/assets/dbd1904b-5f7b-4827-b34e-16cc161a0798

Solid robot is vision simulator ground truth (the "actual" robot pose).
Blue ghost is the drivetrain's estimate of robot position based on vision and odometry.
Red ghost is the latest robot pose seen by the simulated camera.

The green robot ghosts are actually the tag poses seen by the cameras but I had to use robot ghosts because, being pose2d, they were hidden under the speaker when using cones or axes.

jkleiber:
@aidnem the verification evidence is great! Could you please create a ticket specifically to analyze and mitigate what is happening at 9 seconds in that video?

I think that we are seeing one red speaker tag and then getting rekt by pose ambiguity, causing the blue ghost to fly off the field. I don't think this is likely in practice since we would be blocked by the stage, but I'd like for us to iron out these issues in sim so we can isolate issues to photonvision/beelink when we use that

aidnem:
Sure, I've created #41.
Also, this should be marginally improved right off the bat by having standard deviations above zero. Right now I think it 100% trusts the vision pose so it will gladly throw our estimate off the sides of the field. However, throwing super wrong poses at the pose estimator will still always hurt the localization so we should fix it anyway.

vision/src/main/java/coppercore/vision/CameraIOPhoton.java
```diff
@@ -0,0 +1,136 @@
...
+
+        return new CameraIOPhoton(camera, layout, params.robotToCamera());
+    }
+
+    @Override
+    public void updateInputs(CameraIOInputs inputs) {
+        inputs.connected = camera.isConnected();
+
+        PhotonPipelineResult result = camera.getLatestResult();
+        if (result.getTimestampSeconds() == latestTimestampSeconds) {
```
jkleiber, line 66: should this be <= ?

jkleiber (commented):
I think I only have one minor nitpick, otherwise I'm pretty much ready to approve

vision/src/main/java/coppercore/vision/CameraIOPhoton.java
aidnem: Makes sense to me. Fixed.

vision/src/main/java/coppercore/vision/CameraIO.java
```diff
@@ -0,0 +1,28 @@
...
+import edu.wpi.first.math.geometry.Rotation2d;
+import org.littletonrobotics.junction.AutoLog;
+
+public interface CameraIO {
+
+    @AutoLog
+    public static class CameraIOInputs {
+        public Pose3d latestFieldToRobot = new Pose3d();
+        public double averageTagDistanceM = 0.0;
+        public Rotation2d averageTagYaw = new Rotation2d();
```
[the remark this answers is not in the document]
aidnem: We calculate it all manually. Created #42 👍

jkleiber (approved):
approved

PR #51, opened 2024-11-26
Vision distance

### LINKED ISSUE #41: Fix vision pose estimates teleporting out of the field
**Purpose**
Sometimes vision pose estimates randomly snap to very wrong positions (sometimes out of the field). This phenomenon can be seen in the video in [this comment](https://github.com/team401/coppercore/pull/40#issuecomment-2495552728) at 9 seconds.

This is probably due to pose ambiguity and our pose estimator picking the wrong option.

**Scope**
- [ ] Reject the poses that do this or make sure to always pick the right pose somehow
- [ ] Prove that it works in sim by replicating the behavior from the video, driving around while rotating to test multiple angles of multiple tags from multiple distances to ensure that the poses are consistent and accurate.

### LINKED ISSUE #55: Add standard deviations to vision estimates based on distance
Opening this issue so that it is clear to me what #51 is related to

**Work To Do**
- Compute std. deviation of vision estimates based on number of tags and distance to tags

linglejack06:
**Summary**: add distance based variance and tag based variance according to photon pose estimation GitHub

- remove filter photon pose method as it is now replaced by a variance with max values
- add a 2025 method that can be commented out and replace current way of finding new results when wpilib is upgraded

*Variances*:
- single tags have a standard deviance and multiple tags have another variance with lower values
- deviations are then multiplied by the average distance squared divided by 30 (this is what photon had, teams use different options throughout)
- if distance is greater than 4 AND only one tag then variances are filled with max value (rejected)

Resolves #55 
Resolves #41


aidnem:
@linglejack06 could you drive it around in sim a little bit and screen record to demonstrate it? Also that would verify if this mitigates the awful pose estimates while turning problem.

vision/src/main/java/coppercore/vision/Camera.java
```diff
@@ -39,7 +38,23 @@ public CameraMeasurement getLatestMeasurement() {
...
+            return stdDev;
+        } else {
+            if (numTags > 1) {
+                stdDev = CoreVisionConstants.multiTagStdDev;
+            } else if (numTags == 1 && avgDistanceFromTarget > 4) {
+                return CoreVisionConstants.rejectionStdDev;
+            }
+
+            // distance based variance
+            stdDev = stdDev.times(1 + (Math.pow(avgDistanceFromTarget, 2) / 30));
```
jkleiber, line 56: Can we un-nest the if/else. For example
```
else if (numTags > 1) { ... }
else if (numTags == 1 && avgDistanceFromTarget > 4)
```

Also, can we make some of these magic numbers constants in CoreVisionConstants.java? In particular, 4 should be some vision cutoff constant, and the parameter(s) used in the distance based variance calculation should be made into constants in order to name them something other than "the 30 on line 56"

A comment on why 1 + x^2/30 would also be appreciated lol

vision/src/main/java/coppercore/vision/CameraIOPhoton.java
```diff
@@ -90,33 +87,29 @@ public void updateInputs(CameraIOInputs inputs) {
...
+    // }
 
-    private static double calculateAverageTagDistance(EstimatedRobotPose pose) {
+    private double calculateAverageTagDistance(EstimatedRobotPose pose) {
         double distance = 0.0;
         for (PhotonTrackedTarget target : pose.targetsUsed) {
+            var tagPose = poseEstimator.getFieldTags().getTagPose(target.getFiducialId());
+            if (tagPose.isEmpty()) {
+                continue;
+            }
```
jkleiber, line 107: is it possible for `tagPose` to be empty, but still count in the `targetsUsed` list (thus making the average wrong)?

Basically: should we maintain our own count instead of using `targetsUsed.size()` below?

jkleiber (commented):
first pass feedback + concur with @aidnem that I'd like to see this drive in sim (we might need to tune some of the standard deviations a bit)

vision/src/main/java/coppercore/vision/Camera.java
linglejack06: this is done, unsure why its x^2/30. i think that term can be messed around with depending on how accurate we find our vision to be from a far distance. (ie make denominator smaller to rely on it less)

vision/src/main/java/coppercore/vision/CameraIOPhoton.java
linglejack06: yes ill update that, good catch

linglejack06:
also copied from above, im unsure why specifically x^2 / 30 is added for distance, but i assume it makes it an easy way to adjust the denominator depending on how much we trust farther measurements

jkleiber:
@linglejack06 you could rebase onto/cherry pick my maven branch if you want to use maven local rather than Jitpack

I'm debating just getting that merged and then coming back for the maven central version

linglejack06:
[Uploading Screen Recording 2024-12-02 at 6.39.37 PM.zip…]()

linglejack06:
https://github.com/user-attachments/assets/453a14c4-91b3-4df9-b296-2010642930b8

linglejack06:
no more drifting, looks great in sim. ready for review @jkleiber @aidnem

linglejack06:
also turning in middle of field is fxed with no loss of pose

jkleiber (approved):
Verification LGTM, Approved

linglejack06:
@aidenm can you review this, its waiting on you?

aidnem:
@linglejack06 sure, I thought it was stuck because of the weird review dismissing behavior.

aidnem (approved):

aidnem:
@linglejack06 It's now stuck because spotless is causing build to fail, you could try the solution I linked under #59 . I haven't tried it so I'm not sure if it will work but it's worth a shot.

linglejack06:
@aidenm i went ahead and just added back the build constants until we decide how to fix it definitively

aidnem (approved):

PR #69, opened 2024-12-08
Set up Vision according to advantage kit

### LINKED ISSUE #29: rework vision
Summary: create a useful vision system across years
- read Photon Vision docs thoroughly
- add basic example and test
- if needed, add more advanced standard deviations and pose rejections (after testing basic)

### LINKED ISSUE #42: Add tag pitch and roll to camera inputs
Currently, the camera IO calculates the average tag yaw for each measurement. We should add pitch and roll. These will be useful for seeing what values cause bad vision estimates when watching replays. This way we can improve pose estimation.

**Scope**
- [ ] Calculate average tag pitch
- [ ] Calculate average tag roll

### LINKED ISSUE #63: Use getAllUnreadResults
Currently, VisionIOPhoton uses getLatestResult, which is deprecated and planned for removal. We should replace it with getAllUnreadResults.

A reference for 2025 vision with logging with AdvantageKit can be found in the [2025 AdvantageKit vision template](https://docs.advantagekit.org/getting-started/template-projects/vision-template)

### LINKED ISSUE #70: Add garbage collection according to Advantage Kit template
Summary: templates have a script in build.gradle that helps rio's collect garbage more efficiently, this should allow better ram usage for us in 2025

linglejack06:
- adds all poses to logging
- adds all tags to logging
- simplifies file structure
- adds pose rejection
- updates a few naming conventions (should better match the talon fx swerve though if thats 2025 plan)

Resolves #63 
Resolves #70 
Resolves #29 
Resolves #42


linglejack06:
this is failing build because advantage kit isnt sure how to autolog its records created (tagObservation and poseObservation) . I assume as the 2025 version not in beta is released, or a future beta version, this issue will be fixed. The only workaround I could do is to break apart the observation into individual arrays of doubles, poses, etc. and keep track of the index for each when looking at the inputs

linglejack06:
overall though, this should lead to a small improvement in actual vision. If there is any it would just be from consuming all of the poses that might have been backed up since we last checked, which i could only see happening if our coprocessor was running without any lag. More importantly though, the logging capabilities of this vision system is much better. Once advantage kit fixes tha auto log, we will be able to go back and view each pose observed which should allow us to better tune standard deviations

aidnem:
@linglejack06 does their template do the same thing? And if so, would you mind testing if that builds?

PChild:
@linglejack06 does [this](https://www.chiefdelphi.com/t/advantagekit-2025-log-replay-streamlined/476019/33?u=pchild) match the problem you're having?

linglejack06:
> @linglejack06 does [this](https://www.chiefdelphi.com/t/advantagekit-2025-log-replay-streamlined/476019/33?u=pchild) match the problem you're having? 

@PChild no, my issue says that it can't find a match to create a network tables instance. It searches through all types, string, int, array, list, pose2d,etc. but none match the custom record defined in template

linglejack06:
> @linglejack06 does their template do the same thing? And if so, would you mind testing if that builds?

@aidnem i can test that, I haven't yet

linglejack06:
@jkleiber the build.gradle now follows advantage kit template exactly (other than our spotless config and maven publishing) it still throws same error related to json auto detect visibility.

Do you think it could be related to not having installed the latest wpilib packages to my computer? i.e my wpilib is still 2024. I don't think this is it because I figure it would show an error more closely relating to this like a bad jdk etc. but maybe this is issue?

linglejack06:
if that is the issue, then its worth trying that on the older build.gradle where we were getting the json auto detect visibility issue without using the gradle rio and everything.

linglejack06:
@jkleiber the only thing i am concerned about with using GradleRio is its compatibility with another GradleRio project when imported into robot code, because a different main class is set up under coppercore. Do you envision this being an issue?

jkleiber:
It could be an issue for sure, but I just want to rule it out as the problem first

linglejack06:
@jkleiber I just tested with wpilib 2025 fully installed. Now the compiler sees errors in wpilib_interface monitored subsystem (logger is used so same errors we were seeing with auto detect visibility). I have made the wpilib_interface build.gradle exactly match advantage kit template as well, which has not solved this issue. I am unsure where to go from here, any ideas? should i try going back to the old build.gradle but with the wpilib 2025 fully installed?

jkleiber:
@linglejack06 for now let's go back to the original build gradle technique. Then let's just not use any advantage kit logging features for now to get the code building. 

We could end up with a core vision system that doesn't depend on advantage kit at all (which would honestly be a good thing imo). 

If we are getting errors in Monitored subsystem due to akit logging failures, let's comment logging out for now and see about using WPILib alerts instead for the time being

aidnem:
@linglejack06 It's bizarre to me that MonitoredSubsystem is causing issues with the logger. All it's doing is logging two booleans per monitor.

Also, since there weren't any errors when building the main branch, can we reasonably conclude that the issue either lies somewhere within this branch's changes or on your machine? That could narrow down what to look for.

Would you mind sending the error message so I could try and figure out if I messed up MonitoredSubsystem?

aidnem:
@jkleiber I added a ticket for using alerts with monitors as well just to track that idea: #71

linglejack06:
@jkleiber @aidnem after further pondering, and killing gradle daemons a few times, there are no more advantage kit logging errors. Now the issue just resolves around the importing of monitors into the wpilib_interface, which i think is just an issue with maven that can easily be solved by justin explaining. here is the error: 
Screenshot 2024-12-11 at 9 30 13 AM

linglejack06:
to clarify, vision fully compiles now with the logging features, wpilib_interface only fails compilation due to an issue with the monitors jar. Therefore, the issues revolving around auto detect visibility, logging, and autolog all appear to have been solved by using gradlerio and annotation processor.

linglejack06:
```
	implementation project(":monitors")
	implementation project(":math")
```

here is how i added the monitors project under dependencies of wpilib_interface

jkleiber:
@linglejack06 Considering that this error doesn't exist on main, it's worth figuring out why this is happening. It's probably related to some change on this branch

It would be good to rebase and resolve merge conflicts to see if this is a problem that is on your computer or if it reproduces in CI

linglejack06:
@jkleiber rebased and still getting error on my computer after killing daemon. looks like ci is failing too

linglejack06:
@jkleiber by switching from implementation project(":monitors"), etc. to implementation "io.github.team401.coppercore:monitors:2025.0.0-beta" the issue was resolved. this obviously just grabs it from maven central. Im not sure why this change was needed, as main works perfectly fine with the implementation project variant

jkleiber:
Maybe stick with that solution until you get all the features made and then we'll revisit how to fix this correctly

linglejack06:
okay, all features are made then. could i get @jkleiber or @aidnem to review the current request before I add the doc strings to everything?

settings.gradle
```diff
@@ -12,4 +12,4 @@ plugins {
 
 rootProject.name = 'coppercore'
 
-include('geometry', 'wpilib_interface', 'math', 'controls', "parameter_tools", 'vision', 'monitors')
+include(':geometry', ':wpilib_interface', ':math', ':controls', ":parameter_tools", ':vision', ':monitors')
```
jkleiber, line 15: why did we do this? Could this be the source of the build issues?

vision/build.gradle
```diff
@@ -9,92 +9,146 @@ plugins {
...
 	id "com.vanniktech.maven.publish" version "0.30.0"
+	id "edu.wpi.first.GradleRIO" version "2025.1.1-beta-2"
 }
 
 java {
 	sourceCompatibility = JavaVersion.VERSION_17
 	targetCompatibility = JavaVersion.VERSION_17
 }
 
+def ROBOT_MAIN_CLASS = "coppercore.vision.VisionLocalizer"
```
jkleiber, line 20: since we are trying gradlerio, can you open a PR to show this works in 2025-Robot-Code?

You could publish an alpha release (something like 2025.0.0-alpha) and pull it from maven central to prove the concept

vision/build.gradle
```diff
@@ -9,92 +9,146 @@ plugins {
...
+task(replayWatch, type: JavaExec) {
+	mainClass = "org.littletonrobotics.junction.ReplayWatch"
 	classpath = sourceSets.main.runtimeClasspath
 }
-// TODO: Figure out why advantagekit won't let us build the project
-//compileJava.finalizedBy checkAkitInstall
 
 // Defining my dependencies. In this case, WPILib (+ friends), and vendor libraries.
-// Also defines JUnit 5.
+// Also defines JUnit 4.
```
jkleiber, line 91: isn't this still JUnit 5?

vision/build.gradle
```diff
@@ -9,92 +9,146 @@ plugins {
...
+wpi.sim.addDriverstation()
+
+// Setting up my Jar File. In this case, adding all libraries into the main jar ('fat jar')
+// in order to make them all available at runtime. Also adding the manifest so WPILib
+// knows where to look for our Robot Class.
+jar {
+	from { configurations.runtimeClasspath.collect { it.isDirectory() ? it : zipTree(it) } }
+	from sourceSets.main.allSource
+	manifest edu.wpi.first.gradlerio.GradleRIOPlugin.javaManifest(ROBOT_MAIN_CLASS)
+	duplicatesStrategy = DuplicatesStrategy.INCLUDE
```
jkleiber, line 139: we shouldn't need to do this (but we'll need to verify via integrating in 2025 robot code). The maven publisher should handle making the jar file.

If it's a problem to not have this, we will need to find a creative way to avoid GradleRIO

vision/build.gradle
```diff
@@ -9,92 +9,146 @@ plugins {
...
-// classPackage = "frc.robot"
-// className    = "BuildConstants"
-// dateFormat   = "yyyy-MM-dd HH:mm:ss z"
-// timeZone     = "America/New_York"
-// indent       = "  "
-// }
+// Configure jar and deploy tasks
+deployArtifact.jarTask = jar
+wpi.java.configureExecutableTasks(jar)
+wpi.java.configureTestTasks(test)
```
jkleiber, line 145: similar to above, we shouldn't need this

vision/src/main/java/coppercore/vision/CoreVisionConstants.java
```diff
@@ -20,4 +20,13 @@ public final class CoreVisionConstants {
...
     public static final double distanceFactor = 30.0;
+    public static final double maxZCutoff = 1.0;
+
+    // Standard deviation multipliers for each camera
+    // (Adjust to trust some cameras more than others)
+    public static double[] cameraStdDevFactors =
+            new double[] {
+                1.0, // Camera 0
+                1.0 // Camera 1
+            };
```
jkleiber, line 31: I think this should be robot specific rather than coppercore defined

vision/src/main/java/coppercore/vision/VisionLocalizer.java
```diff
@@ -1,62 +1,176 @@
...
+        }
+
+        logSummaryData(allRobotPoses, allRobotPosesAccepted, allRobotPosesRejected);
+    }
+
+    public void setVisionConsumer(VisionConsumer consumer) {
+        this.consumer = consumer;
+    }
+
+    private boolean rejectPose(VisionIO.PoseObservation observation) {
```
jkleiber, line 98: Since this is a function that returns a boolean, let's name it `shouldRejectPose` or something similar

wpilib_interface/build.gradle
```diff
@@ -5,58 +5,117 @@ import com.vanniktech.maven.publish.JavadocJar
 
 plugins {
 	id "java"
+	id "maven-publish"
```
jkleiber, line 8: we should not have this in the build.gradle now that we have `id "com.vanniktech.maven.publish" version "0.30.0"`

wpilib_interface/build.gradle
```diff
@@ -5,58 +5,117 @@ import com.vanniktech.maven.publish.JavadocJar
...
 	id "com.vanniktech.maven.publish" version "0.30.0"
+	id "edu.wpi.first.GradleRIO" version "2025.1.1-beta-2"
 }
 
 java {
 	sourceCompatibility = JavaVersion.VERSION_17
 	targetCompatibility = JavaVersion.VERSION_17
 }
 
+def ROBOT_MAIN_CLASS = "coppercore.wpilib_interface.DriveWithJoysticks.java"
```
jkleiber, line 20: see other comment about testing this in 2025-Robot-Code

wpilib_interface/build.gradle
```diff
@@ -66,38 +125,35 @@ test {
...
+
+// Setting up my Jar File. In this case, adding all libraries into the main jar ('fat jar')
+// in order to make them all available at runtime. Also adding the manifest so WPILib
+// knows where to look for our Robot Class.
+jar {
+	from { configurations.runtimeClasspath.collect { it.isDirectory() ? it : zipTree(it) } }
+	from sourceSets.main.allSource
+	manifest edu.wpi.first.gradlerio.GradleRIOPlugin.javaManifest(ROBOT_MAIN_CLASS)
+	duplicatesStrategy = DuplicatesStrategy.INCLUDE
 }
```
jkleiber, line 145: This is extremely sus to me - I think this addition is probably causing the dependency issues in CI/locally since the jar is being build out of process (the jar should be handled by mavenPublisher rather than here)

jkleiber (commented):
First pass feedback. I think the vision code looks good (needs to be tested though), so I mostly left comments on build scripts

vision/build.gradle
linglejack06: still builds without this, testing in 2025 soon

wpilib_interface/build.gradle
linglejack06: it was

settings.gradle
linglejack06: this was changed attempting to fix the build issues, this is what gradle docs say to do for subprojects. It made no difference to the build failing or not though. build currently works with this

linglejack06:
@jkleiber after updating high key to 2025 and moving ground truth pose calculation to phoenix drive. vision estimates are great near amp and speaker. However, when going to corners or middle of field, both the ground truth and vision begin to randomly set a rotation. I assume its something wrong with my ground truth pose, as vision has no influence on ground truth and ground truth is still setting random rotation. Any clue as to why this is happening? Would you like me to create a repo for high key 2025 in team 401?

linglejack06:
@jkleiber i just ran simulation with advantage kits vision template exaclty (using their diff drive too) it works perfectly, so i think im gonna try to use their diff drive example in high key 2025 to see if i can isolate issue being vision or phoenix drive. then ill move from there depending on what issue is

jkleiber:
Rather than creating a new repo, I would just make a PR against the current repo that upgrades the project to 2025 + demonstrates vision

linglejack06:
ok i can do that

linglejack06:
@jkleiber here is a demo with the demo drive (diff drive) template used. Oddly, when using my vision, i could only set the max voltage outputted as 1 rather tahn 12, otherwise the robot moved incredibly fast. This was not an issue when using advantagekit demo drive with their demo vision, could this be caused by vision? 

I also noticed that when moving through the center of the field the robot seemed to speed up, i feel like these things are related to my vision estimates, but im not sure how.

Other than that, the new logging features are great as we can see when poses are rejected / accepted (shown to the left of screen looking at length of lists) also average distance updates much faster.

Heres the video: 

https://github.com/user-attachments/assets/475c94e2-7cb7-432a-a9c6-ff01d5d9e868

@aidenm also, do you have any idea why vision might be moving robot very fast?

linglejack06:
looking at logs, it looks like robot periodic is consistently overshooting loop time by about 50 milliseconds, could it be getting backed up and then updating the poses all at once?

jkleiber:
Maybe when we're in the middle of the field we get more measurements and vision becomes non-performant? To see if that's the case you could drive to the middle and sit still to see if loop times and number of tags is considerably higher

Other than the performance issues this looks good. Are you planning on trying this again with swerve?

linglejack06:
Yes for swerve, I was just ruling it out as the issue, which I'm not sure is ruled out yet due to center of field issues. As for swerve, do you want me to attempt to integrate vision with the phoenix drive train. Or just copy over the advantage kit swerve project (since that's what we're using in 2025).

As for vision center of field issues, that could be it, but I'm not sure why that would be doing anything different than advantage kit's example as they are the same other than extraction of methods. @jkleiber

linglejack06:
I'll try sitting in the middle for a while and recording that though to officially rule it out

jkleiber:
If copying over the advantagekit 2025 swerve project is more likely to work, let's do that.

Theoretically @sleepyghost-zzz will get the drivetrain working for 2025-Robot-Code soon and we'll be able to hard cut over to using that repo on highkey before we get comp bot

linglejack06:
I think it will be easier to avoid swerve ground truth issues. I'll do that as it'll also ensure that connection is fully working by the time we integrate vision into 2025-robot-code. I'll get that done soon.

aidnem:
@linglejack06 sorry I didn't see this sooner. Could it be the same issue as before, where bad poses from the middle aren't getting rejected? If the pose estimator is somehow consistent but still accepting bad poses, could it conceivably be receiving a bunch of poses outside of the edge of the field and then blending those into your pose estimate so that it looks like the robot is zooming toward the edge of the field?
Do you know if ground truth also moves super fast or just the pose estimate? If you recorded it with ground truth logged and then estimated pose as a ghost you could watch the difference to see if ground truth is also speeding up near the middle of the field.

linglejack06:
@aidenm using the advantage kit example diff drive there is no ground truth, however, the issue would not be estimates outside of field as those are filtered out already. It also is only accepting middle of field poses when tag distance is low enough. It likely is neither of these issues as when i run the full advantage kit template on their github, none of these issues occur. My vision follows their template and my demo diff drive is the exact code from advantage kit github, thats what is so confusing. I also completely copy pasted their template robot container, so im unsure where the problem lies.

The vision is the only slightly different thing, but all pose rejection and standard deviations are done the same way, just extracted into methods.

linglejack06:
im gonna try simulating with their talon fx swerve template soon and see if i run into the same issues

linglejack06:
@jkleiber @aidenm vision is integrated into high-key with 2025 talon fx swerve template. Everything works properly, poses rejected outside of field, etc. Attached is a video where you can see the list of accepted poses shrinking when leaving field and increasing when entering. 


https://github.com/user-attachments/assets/7fcd7756-a592-452b-9112-d5e8238a5a14


Oddly, the robot is still moving very fast, it appears very sensitive. Im not sure if this is a vision issue or a swerve sim issue. vision seems unlikely as i feel it would be jerkier if that was the case. Any ideas?

linglejack06:
also, the issue where drive seemingly speeds up through the middle of the field doesnt appear to be happening, but its hard to tell with how fast the robot is moving all of the time.

aidnem:
You could log the actual speed of the robot (not x or y but use distance formula to calculate total speed) and then graph it and see if it goes above what it should at any point if you want to definitively rule it out @linglejack06. That sensitivity does look pretty insane, did this also happen when you just added DriveWithJoysticks from coppercore before adding vision?

jkleiber (approved):
Approved - let's get this merged so it can be tested asap on highkey 2025 once the game is revealed and the real WPILib drops for 2025

linglejack06:
@aidenm I've only tested all three combined, haven't tried just drive with joysticks with swerve. Nor have I tried just swerve with vision without the drive with joysticks command. I'll mess around with these options if I have time before we get back, otherwise, I'll see what Justin wants me to do

jkleiber:
@linglejack06 can you make a ticket for this investigation work in 2025-robot-code if one doesn't exist. Just to help track the work

PR #72, opened 2024-12-12
Converts the teeth amount of a gear to the pitch diameter.

Samson560915:


PChild:
Can we add [units](https://docs.wpilib.org/en/latest/docs/software/basic-programming/java-units.html) to all of the methods please?

math/src/test/java/coppercore/math/GearConversionTest.java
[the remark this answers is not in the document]
avidraccoon: To use it just provide a third parameter which it the delta

math/src/test/java/coppercore/math/GearConversionTest.java
```diff
@@ -0,0 +1,84 @@
+package coppercore.math;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class GearConversionTest {
+    private static final double delta = 1 * Math.pow(10.0, -10.0);
```
jkleiber, line 7: This is fine, but for reference you can do:
```
private final double kEpsilon = 1e-10;
```

math/src/test/java/coppercore/math/GearConversionTest.java
```diff
@@ -0,0 +1,84 @@
...
+    @Test
+    public void pitchDiameterFrom3mmPulleyTest() {
+        Assertions.assertEquals(
+                12 / (Math.PI), GearConversionFunctions.pitchDiameterFrom3mmPulley(4), delta);
+        Assertions.assertEquals(
+                150 / (Math.PI), GearConversionFunctions.pitchDiameterFrom3mmPulley(50), delta);
+        Assertions.assertEquals(
+                0 / (Math.PI), GearConversionFunctions.pitchDiameterFrom3mmPulley(0), delta);
+        Assertions.assertEquals(
+                21 / (Math.PI), GearConversionFunctions.pitchDiameterFrom3mmPulley(-7), delta);
```
jkleiber, line 18: Let's make this (and other functions) a little cleaner by using arrays and a loop:
```
final double[] kInputs = {4.0, 50.0, 0.0, -7.0};
final double[] kExpectedOutputs = {12.0, 150.0, 0.0, 21.0};

for (int i = 0; i < kInputs.length; i++) {
    Assertions.assertEquals(kExpectedOutputs[i] / (Math.PI), GearConversionFunctions.pitchDiameterFrom3mmPulley(kInputs[i]), delta);
}
```

This way the test inputs and expected outputs are super clear, and the test is set up in a simple way that is easy to read

math/vendordeps/WPILibNewCommands.json
jkleiber, line 1: Is this file actually required? My intuition says no and that it should be deleted. If it is required, it should be upgraded to 2025 though

math/src/main/java/coppercore/math/GearConversionFunctions.java
```diff
@@ -0,0 +1,47 @@
+package coppercore.math;
+
+import static edu.wpi.first.units.Units.*;
+
+public class GearConversionFunctions {
+    public static double pitchDiameterFrom3mmPulley(int teeth) {
```
jkleiber, line 6: These functions likely will fail to compile since we are returning a `Distance` but saying we will return a `double`. We should mark this and other functions to return a `Distance`

jkleiber (commented):
Another round of feedback

avidraccoon:
@jkleiber Why is only WIP running shouldn't the other checks be running?

jkleiber:
@avidraccoon CI doesn't run when there are merge conflicts

avidraccoon:
I was just wondering because the only time it has run on this was the first commit.

jkleiber:
@Samson560915 I want to get this change merged in by early next week - will that be possible?

Samson560915:
I think I did this so can I merge this.

jkleiber (approved):

PR #77, opened 2025-01-06
add javadocs to vision classes

linglejack06:
- javadocs now added to all vision methods
- explanations regarding difficult methods embedded within code

this is updated from josh' vision docs due to migration to advantage kit vision template


jkleiber (approved):
looks good

PR #90, opened 2025-01-17
add 2025 photon vision

linglejack06:
- add 2025 photon vendor dep
- adjust things from beta

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -12,4 +12,4 @@ plugins {
 
 rootProject.name = 'coppercore'
 
-include('geometry', 'wpilib_interface', 'math', 'controls', "parameter_tools", 'monitors', 'vision') // Vision isn't published until photonvision 2025.0.0 is released
+include('geometry', 'wpilib_interface', 'math', 'controls', "parameter_tools", 'monitors', 'vision')
--- a/settings.gradle
+++ b/settings.gradle
@@ -12,4 +12,4 @@ plugins {
 
 rootProject.name = 'coppercore'
 
-include('geometry', 'wpilib_interface', 'math', 'controls', "parameter_tools", 'monitors') // Vision isn't published until photonvision 2025.0.0 is released
+include('geometry', 'wpilib_interface', 'math', 'controls', "parameter_tools", 'monitors', 'vision') // Vision isn't published until photonvision 2025.0.0 is released
--- a/vision/src/main/java/coppercore/vision/VisionIO.java
+++ b/vision/src/main/java/coppercore/vision/VisionIO.java
@@ -1,13 +1,13 @@
 package coppercore.vision;
 
+import edu.wpi.first.apriltag.AprilTagFieldLayout;
 import edu.wpi.first.math.geometry.Pose3d;
 import edu.wpi.first.math.geometry.Rotation2d;
 import org.littletonrobotics.junction.AutoLog;
 
-
 /**
- * Represents an IO wrapper for vision localizer to query. 
- * This class is responsible for updating camera poses and data
+ * Represents an IO wrapper for vision localizer to query. This class is responsible for updating
+ * camera poses and data
  */
 public interface VisionIO {
     @AutoLog
@@ -32,4 +32,6 @@ public interface VisionIO {
             double averageTagDistance) {}
 
     public default void updateInputs(VisionIOInputs inputs) {}
+
+    public default void setAprilTaglayout(AprilTagFieldLayout tagLayout) {}
 }
--- a/vision/src/main/java/coppercore/vision/VisionIO.java
+++ b/vision/src/main/java/coppercore/vision/VisionIO.java
@@ -33,5 +33,5 @@ public interface VisionIO {
 
     public default void updateInputs(VisionIOInputs inputs) {}
 
-    public default void setAprilTaglayout(AprilTagFieldLayout tagLayout) {}
+    public default void setAprilTagLayout(AprilTagFieldLayout tagLayout) {}
 }
--- a/vision/build.gradle
+++ b/vision/build.gradle
@@ -29,13 +29,12 @@ repositories {
 		url "https://frcmaven.wpi.edu/artifactory/littletonrobotics-mvn-release/"
 	}
 	mavenLocal()
-	mavenCentral()
 }
 
-task(replayWatch, type: JavaExec) {
-	mainClass = "org.littletonrobotics.junction.ReplayWatch"
-	classpath = sourceSets.main.runtimeClasspath
-}
+// task(replayWatch, type: JavaExec) {
+// 	mainClass = "org.littletonrobotics.junction.ReplayWatch"
+// 	classpath = sourceSets.main.runtimeClasspath
+// }
 
 // Defining my dependencies. In this case, WPILib (+ friends), and vendor libraries.
 // Also defines JUnit 5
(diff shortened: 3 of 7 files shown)


linglejack06:
@jkleiber @aidnem photon 2025 release finally, just have to fix some build errors relating to other imports and probably beta changes, then this will be ready for a coppercore release.

linglejack06:
@jkleiber for some reason, none of the imports are resolving in vision, do you know why? build.gradle is the same as wpilib_interface?

jkleiber:
@linglejack06 is it failing to resolve in the code editor, or at compile?

If the former, there might just be syntax errors confusing intellisense / needing to refresh after installing a new dependency. 

If the latter, I'd recommend taking a closer look at the build.gradle to make sure the required packages are being included. You could also try killing the gradle daemon and building clean

linglejack06:
> @linglejack06 is it failing to resolve in the code editor, or at compile?
> 
> If the former, there might just be syntax errors confusing intellisense / needing to refresh after installing a new dependency.
> 
> If the latter, I'd recommend taking a closer look at the build.gradle to make sure the required packages are being included. You could also try killing the gradle daemon and building clean

it was failing at compile due to several code errors, which in turn was crashing intellisense (making me think it was import errors). its all fixed now, and should be ready for review.

i switched to using photon utils for single tag estimation, as it was more concise.
went from getting multiple transforms to just calling the photon util estimate pose method which should still work the same. 

I mention as this is only meaningful code change for review, otherwise its just updat to 2025.

linglejack06:
@jkleiber should this release be 2025.0.3 or 2025.1.0 (since there are larger changes to vision with single tag?)

jkleiber:
@linglejack06 maybe we should just switch to release date format (i.e. 2025.1.17 if released today)

aidnem:
@jkleiber @linglejack06 we should definitely do 2025.1.0 if we don't switch to date format. My only issue for dating releases with dates is that it might take more than 1 day to review a PR and by that time we might publish releases whose versions are a few days old.

aidnem:
Also @linglejack06 do we need to use gradleRIO or can you just import wpilib and photonlib like we did [here](https://github.com/team401/coppercore/blob/ae92bd015cfea62f3e15e05ecaa129c58aa316b1/vision/build.gradle)?

The imports would look something like:
```groovy
	def wpilibVersion = "2025.1.1"
	implementation "edu.wpi.first.wpilibj:wpilibj-java:$wpilibVersion"
	implementation "edu.wpi.first.wpilibNewCommands:wpilibNewCommands-java:$wpilibVersion"
	implementation "edu.wpi.first.apriltag:apriltag-java:$wpilibVersion"

	//def akitJson = new groovy.json.JsonSlurper().parseText(new File(projectDir.getAbsolutePath() + "/vendordeps/AdvantageKit.json").text)
	//annotationProcessor "org.littletonrobotics.akit.junction:junction-autolog:$akitJson.version"
	def akitVersion = "4.0.0"
	annotationProcessor "org.littletonrobotics.akit.junction:junction-autolog:$akitVersion"

	implementation "org.littletonrobotics.akit.junction:junction-core:$akitVersion"

	def photonVersion = "v2025.0.0"
	implementation "org.photonvision:photonlib-java:$photonVersion"
	implementation "org.photonvision:photontargeting-java:$photonVersion"

```
This is from an earlier commit that passed build so I assume it worked at the time.

jkleiber:
For releases, we can just make a quick PR to set the version number on the date of release as needed. That way feature PRs don't need to update the version.

Also it makes it more obvious in the commit history where a release was done

linglejack06:
> For releases, we can just make a quick PR to set the version number on the date of release as needed. That way feature PRs don't need to update the version.
> 
> Also it makes it more obvious in the commit history where a release was done

that could work too, im fine with whichever. Just wanted to know to update the release in PR

linglejack06:
> Also @linglejack06 do we need to use gradleRIO or can you just import wpilib and photonlib like we did [here](https://github.com/team401/coppercore/blob/ae92bd015cfea62f3e15e05ecaa129c58aa316b1/vision/build.gradle)?
> 
> The imports would look something like:
> 
> ```groovy
> 	def wpilibVersion = "2025.1.1"
> 	implementation "edu.wpi.first.wpilibj:wpilibj-java:$wpilibVersion"
> 	implementation "edu.wpi.first.wpilibNewCommands:wpilibNewCommands-java:$wpilibVersion"
> 	implementation "edu.wpi.first.apriltag:apriltag-java:$wpilibVersion"
> 
> 	//def akitJson = new groovy.json.JsonSlurper().parseText(new File(projectDir.getAbsolutePath() + "/vendordeps/AdvantageKit.json").text)
> 	//annotationProcessor "org.littletonrobotics.akit.junction:junction-autolog:$akitJson.version"
> 	def akitVersion = "4.0.0"
> 	annotationProcessor "org.littletonrobotics.akit.junction:junction-autolog:$akitVersion"
> 
> 	implementation "org.littletonrobotics.akit.junction:junction-core:$akitVersion"
> 
> 	def photonVersion = "v2025.1.1"
> 	implementation "org.photonvision:photonlib-java:$photonVersion"
> 	implementation "org.photonvision:photontargeting-java:$photonVersion"
> ```
> 
> This is from an earlier commit that passed build so I assume it worked at the time.

@aidenm i can try this

linglejack06:
@aidnem  just tested, it might work but would require several more wpilib imports. do we want to do that @jkleiber or just keep it how it is, i see no issue wiht using gradle unless someone else does

linglejack06:
pretty sure when we tried using the imports we consistently ran into errors due to needing more and more packages, so we decided towards end of december to just use gradle

aidnem:
Okay I see. I suppose it doesn't matter as much if we need all of those dependencies since they'll already be installed in a robot project anyway so (hopefully) they shouldn't cause build times to increase.

jkleiber:
@linglejack06 I'm fine with either approach since it's an external dependency

linglejack06:
@jkleiber  Ok I'll keep current implementation, this is ready for review then I'll publish a release pr with whatever scheme we decide.

Hopefully Saturday or early next week we can start calibrating cameras.

I assume workflow will still be drive works -> basic otf and alignment -> begin incorporating vision to improve otf and autos

vision/src/main/java/coppercore/vision/VisionIO.java
```diff
@@ -32,4 +32,6 @@ public static record PoseObservation(
             double averageTagDistance) {}
 
     public default void updateInputs(VisionIOInputs inputs) {}
+
+    public default void setAprilTaglayout(AprilTagFieldLayout tagLayout) {}
```
jkleiber, line 36: Likely need to add javadoc strings here to avoid making Samson write even more

Also we should fix the camel case: `setAprilTagLayout`

settings.gradle
```diff
@@ -12,4 +12,4 @@ plugins {
 
 rootProject.name = 'coppercore'
 
-include('geometry', 'wpilib_interface', 'math', 'controls', "parameter_tools", 'monitors') // Vision isn't published until photonvision 2025.0.0 is released
+include('geometry', 'wpilib_interface', 'math', 'controls', "parameter_tools", 'monitors', 'vision') // Vision isn't published until photonvision 2025.0.0 is released
```
jkleiber, line 15: Comment is outdated

jkleiber (commented):
two small nitpicks, after that good to go

vision/src/main/java/coppercore/vision/VisionIO.java
linglejack06: I'll add the doc strings, didn't even notice layout wasn't capitalized

jkleiber (approved):

PR #101, opened 2025-01-23
Add path provider for multi-platform support

### LINKED ISSUE #102: Fix CI using deprecated artifact actions v3
Currently CI uses upload-artifact v3, which is deprecated, as described [here](https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/). This is causing build CI to fail on some branches (it auto-fails to warn of the deprecation).

**Scope**
- [ ] Update CI to not use artifact actions v3

aidnem:
Adds environment handler and path provider so that you can use multiple sets of constants depending on what platform you're on

98% of this code is @avidraccoon 's but we had to make some last minute bug fixes, so this branch exists because the commit history diverged by accident during the bugfix process.

Closes #102 by updating to upload-artifacts v4


aidnem:
Wow turns out in the branching process it somehow didn't actually keep any of the bugfix commits I made.
Luckily I remember them pretty well but I'll fix and then test to make sure they work.

jkleiber:
@aidnem if you increment the upload-artifacts to v4 in this branch it'll pass CI and I'll approve

aidnem:
@jkleiber done!

jkleiber (approved):

PR #104, opened 2025-01-24
Vision lineup

linglejack06:
- add cross track distance error method
- add along track distance error method
- add single tag observation record for helping with observations


linglejack06:
@jkleiber could you check over my math calculations in the methods in vision localizer?

also... for the cross track error, how will we determine left/right? how its calculated now would always be a positive number i think? unless target yaw goes negative when camera is to the right of target?

vision/src/main/java/coppercore/vision/VisionLocalizer.java
```diff
@@ -64,6 +66,72 @@ public VisionLocalizer(
...
+        SingleTagObservation tagObserved = inputs[desiredCameraIndex].latestSingleTagObservation;
+
+        // if tag id doesn't match, we assume we don't have that tag in view
+        // therefore, no distance can be observed
+        if (tagObserved.tagId() != tagId) {
+            return -1;
+        }
+
+        // we can subtract the angle of camera if there is one
+        return tagObserved.distance3D() * Math.sin(tagObserved.tx().minus(new Rotation2d()).getRadians())
```
aidnem, line 102: Why does this .minus(new Rotation2d())? Wouldn't that just be zero?

linglejack06: Yes @aidnem I added it as an example. We would need a rotation 2d angle there in cases where our camera is rotated inwards or outwards / upwards or downwards

So the new rotetion2d is a placeholder to explain that that might be needed

vision/src/main/java/coppercore/vision/VisionLocalizer.java
```diff
@@ -64,6 +66,72 @@ public VisionLocalizer(
...
+     *
+     * @param tagId desired tag to align to
+     * @param desiredCameraIndex camera to use for measurements
+     * @param crossTrackOffsetmeters how much to offset horizontal distance by
+     * @return cross track error from tag + offset
+     */
+    public double getCrossTrackDistanceErrorToTag(
+            int tagId, int desiredCameraIndex, double crossTrackOffsetMeters) {
+        // camera not in vision
+        if (desiredCameraIndex >= inputs.length) {
```
jkleiber, line 89: Rather than return -1 here or below, we should do one of the following things:

1. Return a huge number and then only accept values within a certain error range (otherwise we do nothing or use the last value or something)

2. Make a new data type with value and validity as fields (possible coppercore future feature)


You can pick, idc which we choose

vision/src/main/java/coppercore/vision/VisionLocalizer.java
jkleiber, line 102: What is tx()?

Could we make this function name more obvious? As the math is written, this has to be an angle to the tag from the camera. We can probably come up with something better (headingToTag ?)

vision/src/main/java/coppercore/vision/VisionLocalizer.java
```diff
@@ -64,6 +66,72 @@ public VisionLocalizer(
...
+        SingleTagObservation tagObserved = inputs[desiredCameraIndex].latestSingleTagObservation;
+
+        // if tag id doesn't match, we assume we don't have that tag in view
+        // therefore, no distance can be observed
+        if (tagObserved.tagId() != tagId) {
+            return -1;
+        }
+
+        // we can subtract the angle of camera if there is one
+        return tagObserved.distance3D() * Math.cos(tagObserved.ty().minus(new Rotation2d()).getRadians())
```
jkleiber, line 131: What is ty?

Since we're doing cos() here I think we can use the same angle as used above.

My concern is that tx and ty aren't actually angles. In which case we should figure out what they are

vision/src/main/java/coppercore/vision/VisionLocalizer.java
jkleiber: Actually okay so ty is vertical angle to tag and tx is yaw angle to tag based on my vague/weak understanding of this. If that's the case wouldn't taking cos of either tx / ty not be quite accurate?

To demonstrate this, consider a scenario where the tag is dead center on the camera in X (tx=0) but ty=45 degrees. Then:
- using tx only: dist*cos(0) is incorrect because we leave out the vertical angle
- using ty only: dist*cos(45 deg) is correct because we factor in the angle

But if we are dead center vertically but the tag is 45 degrees to one side:
- using tx only: dist*cos(45 deg) is right because we have no vertical error but do have a yaw error 
- using ty only: dist*cos(0) is wrong as this is just the hypotenuse

Another way to think of it: The distance to the tag is represented by a line in 3d space, so both angles matter.

This almost certainly impacts the crosstrack error equation as well. I am on my phone so deriving this is too tough rn but I'll think about it some more. Give it a shot though

jkleiber (commented):
Math feedback - I'll be at shop starting from around 12 and we can try working this out

jkleiber:
I think these are the correct equations:

xtk error = 3d_dist * cos(vertical angle) * sin(yaw angle)

forward distance= 3d_dist * cos(vertical angle) * cos(yaw angle)

If yaw angle is signed then we'll have a signed xtk error.

vision/src/main/java/coppercore/vision/VisionLocalizer.java
linglejack06: Ok I can probably just make another record

linglejack06:
@jkleiber those equations probably make more sense, for forward distance though, if we take cos of vertical and cos of horizontal, aren't we then getting to forward distances away? (For times when one isn't 0?) I might just need to work it out on paper to make exact sense of your equations though as I'm sure they work

linglejack06:
Is this similar to your equation: [3d position line](https://stackoverflow.com/questions/32551940/find-a-new-3d-position-from-a-known-point-of-origin-based-on-distance-and-angle)

jkleiber:
@linglejack06 I'm not sure what you mean by forward distances away. But my goal with the forward distance calculation is to get the distance between the robot and the reef side along the vector perpendicular to the reef (along-track distance)

In my forward distance equation, it can only be 0 if 3d_dist is 0 (since we won't be able to observe the tag at 90 degrees). For the xtk error, it'll be 0 if 3d_dist is 0 or if we are centered. So we shouldn't modify the 3d_dist measurement to correct for camera position before computing xtk error (we should apply any offset afterwards)

It does look like the equations I posted are similar to radius-azimuth-elevation, I just defined the angles slightly different

linglejack06:
That’s ridiculous lolSent from my iPhoneOn Jan 25, 2025, at 4:29 PM, Justin Kleiber ***@***.***> wrote:﻿
@linglejack06 shortly after you left we got the JSON constants loading, and I just pushed to the test-integration branch
@avidraccoon identified the problem as being the need for using Double rather than double.

linglejack06:
@jkleiber i updated calculations and switched to just one method that returns a record.

before we test and merge this, do you think its worthwhile to check if the tag observation is stale, ie if timestamp is too far behind robot timestamp, if so, would this just be passed into the method i assume (latest timestamp)

also... right now single tag observations are only gathered if there is no multitag result. I could also gether them by using the best target in the multitag result. This might be helpful if we are far to where another reef side tag accidentally gets into the frame.

in which case, the robot would recognize a multitag result, but we would still want the ability to use the best target (which should be the tag we are viewing straight on) to use our landing zone alignment?

linglejack06:
let me know if my explanations regarding multitag result doesn't make sense, i dont think i worded it that well

jkleiber:
@linglejack06 I think if we're far enough away to see multiple tags we should be using global pose in any lineup logic

```
if (isMultitag) {
    // Compute offset distances from global pose
} else {
    // Compute offset via tag relative math
}
```

I think that logic would live in the drivetrain code though

linglejack06:
> @linglejack06 I think if we're far enough away to see multiple tags we should be using global pose in any lineup logic
> 
> ```
> if (isMultitag) {
>     // Compute offset distances from global pose
> } else {
>     // Compute offset via tag relative math
> }
> ```
> 
> I think that logic would live in the drivetrain code though

ok, thats a good option, the multitag would be realyl easy too as id just use the pose gathered in drive with vision

linglejack06:
@jkleiber this should be ready for review

jkleiber (approved):
Approved

PR #108, opened 2025-01-30
Vision single tag observations (in multitag results)

linglejack06:
- add hasMultitagRessult method to VisionLocalizer
- update inputs in favor of an array of single tag observations
- update distanceToTag method to utilize the array, interfacing with multitag results as well for single tag lineup


vision/src/main/java/coppercore/vision/VisionLocalizer.java
```diff
@@ -75,6 +75,17 @@ public Rotation2d getTargetX(int cameraIndex) {
         return inputs[cameraIndex].latestTargetObservation.tx();
     }
 
+    public boolean hasMultitagResult() {
+        boolean hasResult = false;
+        for (VisionIOInputsAutoLogged input : inputs) {
+            if (input.hasMultitagResult) {
+                hasResult = true;
```
jkleiber, line 82: you could break out of the loop here

jkleiber (approved):
approved - my only comment is a micro-optimization so feel free to not fix it (or I can approve quickly if you do choose to fix it)

linglejack06:
@jkleiber updated for optimization

PR #111, opened 2025-01-31
Fix missing transition actions

avidraccoon:


avidraccoon:
@aidnem did this work for you?

aidnem:
Yes it worked for Claw state machine

controls/src/main/java/coppercore/controls/state_machine/StateMachine.java
```diff
@@ -51,27 +51,33 @@ public void fire(Trigger trigger) {
...
-                } else {
+                } else if (config.hasEntryAction()) {
                     config.runOnEntry(transition);
+                } else {
+                    runOnEntry(transition);
                 }
             } else {
                 configuration.runOnEntry(transition);
             }
+        } else {
```
aidnem, line 78: Is this if-statement just so that we don't run onEntry on an internal transition?
Also what is an internal transition, is that when we transition from a state to itself or something?

aidnem (commented):
Everything looks good, just 1 question

controls/src/main/java/coppercore/controls/state_machine/StateMachine.java
avidraccoon: It is when we don't want to run on Entry or on Exit

aidnem: But how do you actually do an internal transition

avidraccoon: What to you mean?

avidraccoon:
It also fixes the onEntry problem

controls/src/main/java/coppercore/controls/state_machine/state/StateConfiguration.java
```diff
@@ -124,7 +132,7 @@ private Optional<Transition<State, Trigger>> filterTransitions(
                 if (conditinal
                         && ((ConditinalTransition<State, Trigger>) transition).isCheckTrue()) {
                     return Optional.empty();
-                } else {
+                } else if (((ConditinalTransition<State, Trigger>) transition).isCheckTrue()) {
                     returnOptional = Optional.of(transition);
                     conditinal = true;
                 }
```
jkleiber, line 138: @avidraccoon please make a ticket to document the logic in the state machine with comments. It takes too long to understand what this code does without a couple high-level explanation comments

My understanding of this function is:
1. Loop through all the transitions
2. If any conditional transitions are true, return the first one.
3. If no conditional transitions are true, but an unconditional transition is true, return the last unconditional transition we find.

Just from writing that comment, I have questions about if we are actually enforcing one unconditional transition per trigger per state. However, I think that should be fixed (if needed) in a follow on issue since this PR is currently blocking critical path development work

jkleiber (approved):
Approving this to unblock @linglejack06 's state machine development. @avidraccoon please create deferred tickets as requested

@linglejack06 please test on this branch, and if it works feel free to merge

linglejack06:
okay, ill use mavenLocal with this branch, and if it works at shop tonight
ill merge

controls/src/main/java/coppercore/controls/state_machine/state/StateConfiguration.java
avidraccoon: It returns no transition if two or more conditional transitions are true

avidraccoon: I need to make it so that permit does not check if conditional permits exist

linglejack06:
@avicraccoon, lineup state now transitions from otf, but still not working
for joystick -> lineup

avidraccoon:
Is the transition still null?

linglejack06:
> Is the transition still null?

@avidraccoon unsure, I'll test more at shop tonight, might just be I have a bad conditional, but it just  stays in joystick if I'm near the final pose, whereas it should go to lineup

jkleiber (approved):
please merge to unblock development. We can make improvements in a new ticket

PR #114, opened 2025-02-03
Vision Lineup fixes

linglejack06:
- switch offsets from addition to subtraction
- fix rotation 2d being passed degrees


jkleiber (approved):
Approved

PR #117, opened 2025-02-05
Add Tunable, TuneG, TuneS, and TuneV

aidnem:
Import the Tunable interface, and then TuneG, TuneS, and TuneV commands from last year to be used for automatic characterization this year.


aidnem:
@jkleiber do you think I should make new versions of these for FOC or just make 'runVoltage' and such actually run it with a current in an FOC based system

linglejack06 (commented):
lgtm, but unsure of exact logic

aidnem:
No idea why its not letting your review count lol

linglejack06:
I'll review again once I get by my computer

aidnem:
We could also use mavenlocal for this and then approve it once we know it works, or I could verify it in sim

aidnem:
@jkleiber could we try to get this merged sometime soon-ish so that the tunable elevator branch can get merged? TuneG on here doesn't really work and I'm not sure why, but I feel like having Tunable in coppercore with only some commands is better than delaying and waiting for all of them.

jkleiber (approved):
Approved to unblock development. @aidnem please make a follow on ticket for making these commands more generic (instead of just being for Angle)

aidnem:
Thanks!

PR #121, opened 2025-02-12
Improved PathProvider

avidraccoon:
Make it so we dont need to tell it which files are in the enviroment and fixes error in JSONSync related to PathProvider


aidnem:
@avidraccoon could you quickly verify that this works with 2025-Robot-Code

avidraccoon:
@aidnem we will probably need to update the config.json file

aidnem:
Yes you should probably try that out and make sure it works before we ship this

avidraccoon:
@aidnem It works on 2025-Robot-Code on branch 37-basic-intake-functionality

gradle.properties
```diff
@@ -1 +1 @@
-version=2025.1.9
\ No newline at end of file
+version=2025.1.10-test
```
jkleiber, line 1: leaving this note to remember to change version string before merge

parameter_tools/src/main/java/coppercore/parameter_tools/path_provider/EnvironmentPathProvider.java
```diff
@@ -26,25 +26,12 @@ public String resolveReadPath(String file) {
...
-        }
-
-        throw new RuntimeException(
-                "Could not find "
-                        + file
-                        + " in the "
-                        + environment.getName()
-                        + " environment at path "
-                        + pathAttempt);
+        return path;
```
jkleiber, line 35: before, we used to throw a RuntimeException for not being able to find a file. Why have we changed this behavior?

Reading this code, I think we will return an incorrect path from this function even if the file doesn't exist. What are the downstream impacts of doing this? Should we restore the original behavior?

parameter_tools/src/main/java/coppercore/parameter_tools/path_provider/Environment.java
```diff
@@ -13,12 +14,8 @@ public String getPath() {
...
 
-    public boolean hasFile(String file) {
-        for (int i = 0; i < files.length; i++) {
-            if (files[i].compareTo(file) == 0) {
-                return true;
-            }
-        }
-        return false;
+    public boolean hasFile(String basePath, String file) {
+        File f = new File(basePath + filepath + file);
```
jkleiber, line 18: Does this not need path separators?

jkleiber (commented):
First pass.

Since this feature is so critical to the rest of the team, there needs to be more verification put into this. I would expect a unit test for each of the following scenarios:
- when a file is found
- when a default is used for a file that cannot be found
- when no file can be found

parameter_tools/src/main/java/coppercore/parameter_tools/path_provider/EnvironmentPathProvider.java
avidraccoon: Do we want to write to the current environment or default, when there the file is not in default or the current environment?

jkleiber: probably the current environment

jkleiber: I think I see what you're trying to do here, but we need a comment to explain why this is okay

My understanding is that you're basically depending on this function in the read/write data functions, and if a file cannot be found this will just return the default. Whereas before we would just fail.

In the case of reading, if there's no path, then we want to fall back on the default. As a last resort we fail out if no default file can be found either. However for writing it is more intuitive that you would not need the file to exist in order to write to it.

So in addition to a comment, we probably need to set up an optional argument (by method overloading) to tell this function if it should fall back on default or not, regardless of if it has a default configured

parameter_tools/src/test/java/coppercore/parameter_tools/test/PathProviderTestData.java
jkleiber, line 1: This class is just a wrapper for `String` - can we remove this and just use String directly?

jkleiber (commented):
Just a couple small changes and then we'll be good

avidraccoon:
Just a note I did add a feature that makes it so we don't need to add the JSONSync in the constants files

parameter_tools/src/main/java/coppercore/parameter_tools/path_provider/EnvironmentPathProvider.java
```diff
@@ -13,9 +13,19 @@ public EnvironmentPathProvider(Environment environment, String filepath, String
...
+    /**
+     * Find path for file for the current enviroment. If File does not exist then it returns the
+     * desired location for the file to be made in.
+     */
+    @Override
+    public String resolveWritePath(String file) {
+        return resolvePath(file, environment.getDefaultToEnvironment());
+    }
+
+    /** Gets path to file This will throw an error if the file does not exist */
```
jkleiber, line 25: this doesn't seem to throw an error if the file doesn't exist - let's update the comment or the code accordingly

parameter_tools/src/test/java/coppercore/parameter_tools/test/PathProviderTests.java
```diff
@@ -0,0 +1,134 @@
...
+
+    @Test
+    public void getWritePathDefualtingMissingTest() {
+        Assertions.assertEquals(
+                pathProvider.getFullPath()
+                        + File.separator
+                        + "defaults"
+                        + File.separator
+                        + "RandomNoneExistantFile.json",
+                testPathProvider.resolveWritePath("RandomNoneExistantFile.json"));
```
jkleiber, line 106: RandomNonExistentFile (spelling)

this also appears in other parts of this file

parameter_tools/src/test/resources/PathProviderTests/config.json
```diff
@@ -0,0 +1,21 @@
...
+      {
+        "name": "normal"
+      },
+      {
+        "name": "custom_path",
+        "filepath": "custom"
+      },
+      {
+        "name": "defaulting",
+        "defaultToEnvironment": true
```
jkleiber, line 17: I don't think we need this to be configurable - I think this is a read vs. write thing. See my other comment in EnvironmentPathProvider.java

If we are reading, we fall back into the default environment.
If we are writing, we don't care if the file exists (we can just write and create the new file)

jkleiber (commented):
@avidraccoon left some more comments, however, we should proceed with the higher priority code in 2025-Robot-Code for now

I'll come back to this PR once the intake and 3D simulation features are complete

parameter_tools/src/test/resources/PathProviderTests/config.json
avidraccoon: Okay that makes more sense

avidraccoon:
I think I should probably install a spell checker

aidnem:
@avidraccoon I've been using [Code Spell Checker](https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker) for a couple months and so a few of our projects have our special words added to dictionary under .vscode already

parameter_tools/src/test/java/coppercore/parameter_tools/test/PathProviderTests.java
```diff
@@ -0,0 +1,136 @@
+/*
+ * Click nbfs://nbhost/SystemFileSystem/Templates/Licenses/license-default.txt to change this license
+ * Click nbfs://nbhost/SystemFileSystem/Templates/UnitTests/JUnit5Suite.java to edit this template
```
jkleiber, line 3: What is this

parameter_tools/src/main/java/coppercore/parameter_tools/path_provider/EnvironmentPathProvider.java
```diff
@@ -24,27 +39,37 @@ public String resolveReadPath(String file) {
...
+            }
         }
 
-        throw new RuntimeException(
-                "Could not find "
-                        + file
-                        + " in the "
-                        + environment.getName()
-                        + " environment at path "
-                        + pathAttempt);
```
jkleiber, line 48: Shouldn't we at least log that we failed to load the file (and are falling back to Java defaults)?

parameter_tools/src/test/resources/PathProviderTests/config.json
jkleiber: I still think we shouldn't need to configure this

jkleiber (commented):
Sort of a half pass here. This PR is fairly old so I don't remember what we even were doing before. 

I think if this was demonstrated in robot code (possibly on hardware) I'd be more at ease. This one really needs a proof of concept

parameter_tools/src/main/java/coppercore/parameter_tools/path_provider/EnvironmentPathProvider.java
avidraccoon: That is done it the resolveRead path because we don't want to throw an error when reading

parameter_tools/src/test/resources/PathProviderTests/config.json
avidraccoon: Is just allow a way to force read and write to use the same directory. It also allows for making more isolated environments

avidraccoon:
I believe more unit tests need to be written.
@aidnem What do you think?

aidnem:
@avidraccoon are you aware of any functionality that isn't currently tested?

The only thing that I see right now that really sticks out is JSONSyncSaveFileTest: it doesn't look like it verifies the data that was written. I would like to see some sort of verification against an expected file: this would prevent us from missing regressions that break writing JSON.

parameter_tools/src/test/resources/PathProviderTests/config.json
aidnem: > If we are reading, we fall back into the default environment. If we are writing, we don't care if the file exists (we can just write and create the new file)

> Is just allow a way to force read and write to use the same directory. It also allows for making more isolated environments

@avidraccoon I have 2 questions about this:
1. Would you mind providing an example of 1 time where the difference in reading and writing would actually come into play? If you were to disable defaulting and then try to read the file, what would happen? Would code crash (file not found), or would it default to the options specified in the java file? If I understand correctly, the only effect that this would have is either forcing it to default to the java constants or to crash your code. Can you think of a use case in which we would want to force defaulting to the java constants and bypass the default environment?
2. Can you elaborate on what you meant by more isolated environments? Could you provide a specific example of an application of an isolated environment as you've described it?

I think once these questions are answered I'll have a better understanding of the purpose of this flag. Thank you!

aidnem:
@avidraccoon before I review this, would you mind going through all of the comments that remain open and giving a short response to them (or resolving them if they're no longer relevant)? I asked a question above as well. Once the change requests are resolved and build is passing, we can take another look at reviewing.

parameter_tools/src/test/resources/PathProviderTests/config.json
avidraccoon: I have replaced this flag with a new flag isolated, that will make it so if the file is not found it will not use the default folders values.

aidnem (changes requested):
I got to the point of reviewing the tests. Everything else looks very very good, except for the few comments that I left. All of the improvements look great. I'll leave this review for now and then you can work on the comments while I review the tests. Thanks for getting this written!

parameter_tools/src/main/java/coppercore/parameter_tools/json/annotations/JsonType.java
```diff
@@ -5,10 +5,26 @@
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
+/**
+ * Annotation to define a JSON type with subtypes for polymorphic deserialization. {@link
+ * JsonSubtype} should be used to define the subtypes. Use will need to use should be used to
+ * register this with the to use.
```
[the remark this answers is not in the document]
avidraccoon: I was trying to do something with javadocs that caused errors, and I forgot to fix this when I removed it.

parameter_tools/src/main/java/coppercore/parameter_tools/json/JSONHandler.java
```diff
@@ -0,0 +1,126 @@
...
+     * Gets JSONSync Object
+     *
+     * @param <T> type of object JSONSync is built for.
+     * @param blankObject blank version of the target type
+     * @param filename file name to load the object from.
+     * @param pathProvider pathProvider to get path to file
+     * @param config JSONSync config
+     * @return JSONSync Object for the target type
+     */
+    public <T> JSONSync<T> getJsonSync(
```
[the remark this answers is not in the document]
avidraccoon: Because it needs to be able to pick which constructor to call.

parameter_tools/src/main/java/coppercore/parameter_tools/json/annotations/JsonSubtype.java
```diff
@@ -1,9 +1,8 @@
 package coppercore.parameter_tools.json.annotations;
 
 /**
- * Annotation to define a subtype for use with {@link JsonType} for polymorphic deserialization. Use
- * will need to use JSONSyncConfigBuilder should be used to register this with the JSONSyncConfig to
- * use.
+ * Annotation to define a subtype for use with {@link JsonType} for polymorphic deserialization.
+ * Used with JsonType.
```
aidnem: In my opinion, you don't need to say Used with Jsontype if you've already said "for use with JsonType" above.

parameter_tools/src/main/java/coppercore/parameter_tools/json/helpers/JSONConverter.java
```diff
@@ -79,11 +77,14 @@ public static <T> Class<? extends JSONObject<T>> convert(Class<T> clazz)
             if (jsonMap.containsKey(clazz)) {
                 return (Class<? extends JSONObject<T>>) jsonMap.get(clazz);
             }
-            return JSONConverter.tryAdvancedConversions(clazz);
+            Class<?> advancedClazz = JSONConverter.tryAdvancedConversions(clazz);
+            if (advancedClazz != null) {
```
aidnem: Good catch with this null check here 👍

parameter_tools/src/test/java/coppercore/parameter_tools/test/PathProviderTests.java
```diff
@@ -0,0 +1,129 @@
...
+            testPathProvider.resolveReadPath("RandomNoneExistentFile.json");
+            Assertions.fail("Should throw error about file missing");
+        } catch (Exception e) {
+            Assertions.assertEquals(
+                    "RandomNoneExistentFile.json does not exist at "
+                            + pathProvider.getFullPath()
+                            + File.separator
+                            + "defaults"
+                            + File.separator
+                            + "RandomNoneExistentFile.json",
```
aidnem: I believe NoneExistant should be NonExistant here.

parameter_tools/src/main/java/coppercore/parameter_tools/json/JSONHandler.java
```diff
@@ -0,0 +1,126 @@
+package coppercore.parameter_tools.json;
```
godmar: This class has 3 overloaded constructors and 4 overloaded `.getJSONSync` methods that fall back to the (optionally provided) configs and path_providers in the constructor.

I find this API choice extremely confusing. Can this be simplified?

For instance, since you always need a config and a path_provider, simply require that each `JSONHandler` have one, and be done with it. If users need to overwrite either config or path_provider they can make a new JSONHandler, which is a cheap (and record-like class) anyway. In fact, it could be written as a record with methods.

parameter_tools/src/main/java/coppercore/parameter_tools/path_provider/EnvironmentHandler.java
```diff
@@ -2,43 +2,83 @@
...
+     * @throws RuntimeException if the specified environment is not found in the list of
+     *     environments.
+     */
     public EnvironmentPathProvider getEnvironmentPathProvider() {
         Environment env = null;
-        for (int i = 0; i < environments.length; i++) {
-            if (environments[i].getName().compareTo(environment) == 0) {
-                env = environments[i];
+        for (Environment environment1 : environments) {
+            if (environment1.getName().compareTo(environment) == 0) {
```
godmar: You're comparing the result of `getName()` to an `environment`. Perhaps there is a way to name these fields better, perhaps using `environmentName`

parameter_tools/src/test/java/coppercore/parameter_tools/test/PathProviderTests.java
godmar: It's spelled existent, not existant. Not to be confused with extant which is spelled with an a.

godmar (changes requested):
Looks good overall.

- consider if JSONHandler can't be simplified
- check on the eaten exception
- document what "advanced" conversions are - or call the "custom" or "fallback"; document why you apply them in order until the first one that succeeds.

parameter_tools/src/main/java/coppercore/parameter_tools/json/adapters/measure/JSONMeasure.java
```diff
@@ -2,150 +2,250 @@
...
+                Units.VoltsPerMeterPerSecondSquared,
+                "VoltsPerMeterPerSecondSquared",
+                "Volts/Meter/Second^2",
+                "V/(m/s²)",
+                "V/(m/s^2)");
+        JSONMeasure.registerUnit(
+                Units.VoltsPerRadianPerSecond,
+                "VoltsPerRadianPerSecond",
+                "Volts/Radian/Second",
+                "V/(rad/s)");
```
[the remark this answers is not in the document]
avidraccoon: I believe the V(rad/s) is right if you check the internal class file

parameter_tools/src/main/java/coppercore/parameter_tools/json/helpers/JSONConverter.java
```diff
@@ -9,66 +9,111 @@
...
     @SuppressWarnings("unchecked")
-    public static <T> Class<? extends JSONObject<T>> convert(Class<T> clazz) {
+    public static <T> Class<? extends JSONObject<T>> convert(Class<T> clazz)
+            throws ConversionException {
         try {
-            return (Class<? extends JSONObject<T>>) jsonMap.get(clazz);
+            if (jsonMap.containsKey(clazz)) {
+                return (Class<? extends JSONObject<T>>) jsonMap.get(clazz);
+            }
+            Class<?> advancedClazz = JSONConverter.tryAdvancedConversions(clazz);
```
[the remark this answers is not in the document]
avidraccoon: It is a fallback

parameter_tools/src/main/java/coppercore/parameter_tools/path_provider/EnvironmentHandler.java
```diff
@@ -2,43 +2,83 @@
...
+ * Handles the loading and management of environments from a JSON file. Provides methods to retrieve
+ * the current environment and its associated path.
+ *
+ * <p>warning To avoid dsync inside application, the same environment handler should be used across
+ * the application. Unless if it is desired to have different environments for different parts of
+ * the application.
+ */
 public class EnvironmentHandler {
-    static JSONSync<EnvironmentHandler> sync =
+    // TODO: Switch to using JSONHandler
```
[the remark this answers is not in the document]
avidraccoon: This is a potential future change not needed or desired in the pr

aidnem (commented):
This looks good to me. I appreciate the improved comments. I don't want to approve until @godmar also gives it the ok. @avidraccoon can we see some verification of this? (maybe use it in 2025 robot code or Serpens 2025 to tune some sim gain).

parameter_tools/src/test/java/coppercore/parameter_tools/test/PathProviderTests.java
aidnem: I'm going to resolve this since spelling won't inhibit functionality and we're on a tight schedule at this point.

PR #130, opened 2025-02-21
116 add cached signal data type

Samson560915:
I don't know if this is what you want or even if this works but can you tell me.


Samson560915:
I think this works pls review.

math/src/main/java/coppercore/math/CachedData.java
```diff
@@ -0,0 +1,77 @@
...
+
+    // Constructor for WriteCount-based expiration
+    public CachedData(int maxWrites) {
+        this.isTimeBased = false;
+        this.maxWrites = maxWrites;
+        this.writeCount = 0;
+    }
+
+    // Update the value based on the change (for both time and write-based mechanisms)
+    public void update() {
```
jkleiber, line 27: This should be called from read() or isStale(). We don't want the user to have to manually call update

math/src/test/java/coppercore/math/test/CachedDataTest.java
jkleiber, line 1: We need a couple more tests in this file:

1. Reading multiple times with valid values until the cache resets and returns null
2. Doing the same thing with time

jkleiber (commented):
First pass feedback

sleepyghost-zzz:
@jkleiber i saw you wanted me to help samson. how do you want us to implement this. aka what does it do. what would you want us to test with this. 
thanks!
-joshua

jkleiber:
@sleepyghost-zzz The main motivation for this is that the photonvision measurements come in slower than 50Hz (which is the frequency our code runs at). So if we don't cache the measurements then we get erroneously null values and lose track of tags 

There is some code in Lineup state that works around this by caching the value. However I want to make a more general data structure in coppercore so ee don't repeat code throughout. I've left some comments on this PR, feel free to address them yourself or work with @Samson560915 (bias toward action)

avidraccoon:
@jkleiber Do you want me to finish this up?

jkleiber:
@avidraccoon It hasn't been worked on in a week so go for it.

Samson560915:
Please make Josh do this, I've spent forever trying to figure out roborio. I think everything else works.

math/src/main/java/coppercore/math/CachedData.java
```diff
@@ -0,0 +1,77 @@
...
+        if (isTimeBased) {
+            long currentTime = System.currentTimeMillis();
```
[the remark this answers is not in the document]
avidraccoon: @jkleiber I believe it would automaticly convert from what ever the robo rio uses to Milliseconds

math/src/test/java/coppercore/math/test/CachedDataTest.java
```diff
@@ -0,0 +1,114 @@
...
+import org.junit.jupiter.api.Test;
+
+public class CachedDataTest {
+    @Test
+    public void isStaleTest() {
+        CachedData<Integer> data = new CachedData<>(1.0);
+        try {
+            for (int i = 0; i < 7; i++) {
+                Thread.sleep(100);
+                Assertions.assertFalse(data.isStale());
```
aidnem, line 19: Somehow this line caused CI tests to fail, but only on MacOS. You could maybe make this test have a little more margin for lag (or check how much time the test has actually taken in the test and make sure that isStale is consistent with this amount of time.

minhnguyenbhs (approved):
this makes _sense_ to me but i don't know if it's quite perfect according to coach justin's goals? it appears to have been tested thoroughly based on what i saw in shop, so you want to go ahead and commit I'm clicking approve, but you might want to talk with coach justin just to make sure :D

minhnguyenbhs:
oh nvm :(

Samson560915:
@aidnem can you please look at this to merge.

minhnguyenbhs:
sorry I spent all of today trying to fix stuff after merging with main, haven't gotten to intake yet
it's very very slow right now, i don't know how to fix it; you practically can't see what's happening until the action's long over
it can kind of run an auto, badly
what do you think of this picture
Untitled

Samson560915:
@jkleiber Can you check this for approval?

gradle.properties
```diff
@@ -1 +1 @@
-version=2025.1.11
\ No newline at end of file
+version=2025.1.samson
```
jkleiber, line 1: The version should be updated to whatever the next release will be

jkleiber (commented):
just one comment to resolve before we merge this!

Samson560915:
@avidraccoon pls review for merge
@aidnem pls review for merge
@jkleiber pls review for merge

math/src/test/java/coppercore/math/test/CachedDataTest.java
```diff
@@ -0,0 +1,126 @@
...
+        } catch (InterruptedException e) {
+            // TODO Auto-generated catch block
+            e.printStackTrace();
+        } // Room for error in Thread.sleep
+        Assertions.assertTrue(data.isStale());
+    }
+
+    @SuppressWarnings({"rawtypes"})
+    @Test
+    public void readTest() {
```
avidraccoon, line 83: Make it test if after so many reads that read returns null.

avidraccoon (changes requested):

math/src/test/java/coppercore/math/test/CachedDataTest.java
Samson560915: no

math/src/test/java/coppercore/math/test/CachedDataTest.java
```diff
@@ -0,0 +1,126 @@
...
+        Assertions.assertEquals(listData.read(), List.of(1, 0, 0, 0, 1, -1));
+        integerData.write(-100);
+        Assertions.assertEquals(integerData.read(), -100);
+        integerData.write(0);
+        Assertions.assertEquals(integerData.read(), 0);
+        integerData.write(100);
+        for (int i = 0; i < 10; i++) {
+            Assertions.assertEquals(integerData.read(), 100);
+        }
+        Assertions.assertEquals(integerData.read(), null);
```
avidraccoon, line 113: Move test null to separate test

avidraccoon (changes requested):

math/src/main/java/coppercore/math/CachedData.java
```diff
@@ -0,0 +1,68 @@
...
+public class CachedData<Type> {
+
+    private Type value = null;
+    private int readCount = 0;
+    private int maxReads = 5;
+    private double lastUpdateTime = 0.0;
+    private double staleTime = -1.0;
+    private boolean isTimeBased = true;
+
+    // Constructor for Time-based expiration
```
avidraccoon, line 12: Actual javadocs

avidraccoon (changes requested):

math/src/test/java/coppercore/math/test/CachedDataTest.java
```diff
@@ -0,0 +1,141 @@
...
+            try {
+                Thread.sleep(50);
+            } catch (InterruptedException e) {
+                // TODO Auto-generated catch block
+                e.printStackTrace();
+            }
+            Assertions.assertFalse(data.isStale());
+        }
+        try {
+            Thread.sleep(1600);
```
avidraccoon, line 73: Instead of just sleeping extra long, can we look into why MacOS is failing, because it might indicate a problem. If you still it is just MacOS being stupid this might be fine.

math/src/test/java/coppercore/math/test/CachedDataTest.java
```diff
@@ -0,0 +1,141 @@
...
+        } catch (InterruptedException e) {
+            // TODO Auto-generated catch block
+            e.printStackTrace();
+        } // Room for error in Thread.sleep
+        Assertions.assertTrue(data.isStale());
+    }
+
+    @SuppressWarnings({"rawtypes"})
+    @Test
+    public void readTestTrue() {
```
avidraccoon, line 83: Can you add new lines, comments, or split the test up, to make it easier to read.

avidraccoon (dismissed):

math/src/test/java/coppercore/math/test/CachedDataTest.java
avidraccoon: You might want to look into making the tests with timing not multithreaded. Thread.sleep could be doing some weird things behind the scenes.

avidraccoon:
@Samson560915 @j0shuaS Will approve once conflicts fixed

Samson560915:
@avidraccoon Help Help Help Help Help Help Help Help HelP HelP HeLP HeLP HElP HeLP HELp HelP HELP

avidraccoon:
Will have to actually sit down and review this tomorrow or tonight

math/src/main/java/coppercore/math/CachedData.java
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * This class checks if the cache is stale
+     *
+     * @return A boolean for whether the cache is stale or not
+     */
+    // Check if the cache has expired (is stale)
+    public boolean isStale() {
+        if (isTimeBased) {
+            double currentTime = System.currentTimeMillis();
```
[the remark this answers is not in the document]
godmar: My reading of the wpilib documentation also seems to indicate that you should not use `System.currentTimeMillis()` at all.  Instead, the `MathSharedStore.getCurrentTime()` gives you the FPGA time. In simulation/testing, you would advance it via [SimHooks.stepTiming](https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/wpilibj/simulation/SimHooks.html#stepTiming(double)). @aidnem can you confirm this?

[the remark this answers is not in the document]
aidnem: @godmar I agree that using MathSharedStore.getTimestamp() would be better. I haven't seen SimHooks.stepTiming before, and I'm currently doing some research to see if that's what we should use. [The getTimestamp() source code](https://github.wpilib.org/allwpilib/docs/release/java/src-html/edu/wpi/first/math/MathSharedStore.html#line-73) calls [MathShared.getTimestamp() defined on Line 32](https://github.wpilib.org/allwpilib/docs/release/java/src-html/edu/wpi/first/math/MathSharedStore.html#line-32) as fetching the time from WPIUtilJNI. I'm not sure if/how the SimulatorJNI and WPIUtilJNI connect (if at all?), so I think using WPIUtilJNI.setMockTime() would be the correct solution. Please let me know if you disagree. I'll do some more digging on this.

I currently can't make my code run without crashing when I try to load the WPIUtilJNI in my branch `149-asymmetric-debouncer`, so I'm still working on figuring out how to make that work in the library. My current assumption is I'm missing part of the build.gradle that specifies how to load the JNI.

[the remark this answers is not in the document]
godmar: I haven't looked at this deeply, but principally, JNI means native code (that is, code written in the C language). That code must be compiled for a specific architecture. When the code is running on the RoboRIO, this would be a library for the ARM Cortex that's on the RoboRIO.  When running on your laptop, you would need an implementation of that library for x86.

How is it "crashing?"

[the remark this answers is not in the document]
aidnem: @godmar sorry, I wasn't very clear with my description. Whenever I run the unit tests for wpilib_interface on branch `149-asymmetric-debouncer`, the process exits with code 1. The error from gradle just tells me that the test runner failed, it doesn't give me a more specific error message.

My current assumption is that I'm missing a piece of configuration that causes the dynamic library to be built, which then results in it being missing at runtime and causing the program to crash at runtime (hence the exit code 1).

If we can get it to work, `enableMockTime` was the plan for Asymmetric Debouncer's unit tests and (in my opinion) should be used here too.

math/src/main/java/coppercore/math/CachedData.java
```diff
@@ -0,0 +1,94 @@
...
+    private int readCount = 0;
+    private int maxReads = 5;
+    private double lastUpdateTime = 0.0;
+    private double staleTime = -1.0;
+    private boolean isTimeBased = true;
+
+    /**
+     * This is a constructor for the Time-based expiration of the Cached Data.
+     *
+     * @param staleTime This is the amount of time it takes a cache to be stale
```
godmar, line 20: Generally, for a @param we would write `@param staleTime the amount of time it takes ...`
Look at how the [standard JDK Javadoc](https://docs.oracle.com/javase/8/docs/api/java/util/Arrays.html#binarySearch-float:A-float-). is written. Try to mimic it in the future.

math/src/test/java/coppercore/math/test/CachedDataTest.java
```diff
@@ -0,0 +1,159 @@
...
+public class CachedDataTest {
+
+    @BeforeEach
+    @SuppressWarnings("unused")
+    void setUpMockTime() {
+        WPIUtilJNI.enableMockTime();
+        setMockTimeSeconds(0.0);
+    }
+
+    public static void enableMockTime() {}
```
godmar, line 25: Why is there this empty function?

math/src/test/java/coppercore/math/test/CachedDataTest.java
```diff
@@ -0,0 +1,159 @@
...
+
+    public static void setMockTime(long time) {}
```
godmar, line 27: Doesn't appear to be used.

math/src/test/java/coppercore/math/test/CachedDataTest.java
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    @Test
+    public void isStaleTest2() {
+        CachedData<Integer> data = new CachedData<>(5);
+        for (int i = 0; i < 5; i++) {
+            Assertions.assertFalse(data.isStale());
+            data.read();
+        }
+        data.read();
```
godmar, line 64: I would `assertEquals` that `data.read()` actually returns 5.

math/src/test/java/coppercore/math/test/CachedDataTest.java
```diff
@@ -0,0 +1,159 @@
...
+    @Test
+    public void isStaleTest() {
+        enableMockTime();
+        CachedData<Integer> data = new CachedData<>(1.5);
+        double staletimeTest = 0.1;
+        for (int i = 0; i < 7; i++) {
+            setMockTimeSeconds(staletimeTest + 0.1 * i);
+            Assertions.assertFalse(data.isStale());
+        }
+        setMockTimeSeconds(10.0);
```
godmar, line 52: 10.0 is excessive for a 1.5 expiration time. Mock time should be accurate.

math/src/main/java/coppercore/math/CachedData.java
```diff
@@ -0,0 +1,94 @@
...
+        this.staleTime = staleTime * 1000; // Convert seconds to milliseconds
+        this.lastUpdateTime = MathSharedStore.getTimestamp() * 1000;
+    }
+
+    /**
+     * This is a constructor for the readCount-based expiration of the Cached Data.
+     *
+     * @param maxReads This is the maximum amount of reads we can take before the cache is stale
+     */
+    public CachedData(int maxReads) {
```
godmar, line 33: Having an `int` and `double` constructor is a recipe for utter confusion. Do not do this.
Use either units or two separate sub classes.  People will get tripped up by the difference between `new CachedData(3)` and `new CachedData(3.)`.

And, for extra credit on the AP exam, predict the output of this program:
```java
public class CachedData {
    boolean isInt;
    public CachedData(int i) { isInt = true; }
    public CachedData(double d) { isInt = false; }

    public static void main(String []av) {
        System.out.println(new CachedData(0).isInt);
        System.out.println(new CachedData(.0).isInt);
        System.out.println(new CachedData(0L).isInt);
    }
}
```
```

File: math/src/test/java/coppercore/math/test/CachedDataTest.java [hunk 17]
```diff
@@ -0,0 +1,159 @@
...
+        } catch (InterruptedException e) {
+            e.printStackTrace();
+        }
+        Assertions.assertEquals(data.read(), -100);
+    }
+
+    @Test
+    public void readTestNull() {
+        CachedData data = new CachedData(1);
+        try {
```
godmar, line 142: Why is this still here? We know that Thread.sleep() is not reliable in the CI/CD pipeline and expiration isn't based on system time.

math/src/test/java/coppercore/math/test/CachedDataTest.java
```diff
@@ -0,0 +1,159 @@
...
+            Thread.sleep(2000);
+        } catch (InterruptedException e) {
+            e.printStackTrace();
+        }
+        Assertions.assertEquals(data.read(), null);
+    }
+
```
godmar, line 149: It seems that there's a test missing for the case when `write()` is called with `null`.  See my comment above.

avidraccoon:
@Samson560915 @godmar What is the current status of this pull request?

math/src/main/java/coppercore/math/CachedDataTime.java
```diff
@@ -0,0 +1,61 @@
+package coppercore.math;
+
+import edu.wpi.first.math.MathSharedStore;
+
+public class CachedDataTime<Type> {
+
+    private Type value = null;
+    private double lastUpdateTime = 0.0;
+    private double staleTime = -1.0;
```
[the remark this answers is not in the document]
aidnem: Also adding javadoc comments to these fields explaining what they are and their units would be helpful.

math/src/main/java/coppercore/math/CachedDataTime.java
```diff
@@ -0,0 +1,61 @@
...
+     *
+     * @param staleTime the amount of time it takes a cache to be stale
+     */
+    public CachedDataTime(double staleTime) {
+        this.staleTime = staleTime * 1000; // Convert seconds to milliseconds
+        this.lastUpdateTime = MathSharedStore.getTimestamp() * 1000;
+    }
+
+    /**
+     * This method is used to write a new signal value(A new cache).
```
aidnem: This is a little nitpicky, but I might rephrase this to "Write a new signal value to the cache." I'm not sure if "A new cache" is factual here since you're not creating a new object or anything.
I also prefer just commenting on methods with what they do, since starting a method docstring with "This method" feels like a waste of information since the user already knows it's a method. My philosophy is a little different for class javadocs since I feel that a more completely/grammatical summary of what the class does is apt since people will likely read that docstring more thoroughly but fewer times.

math/src/main/java/coppercore/math/CachedDataTime.java
```diff
@@ -0,0 +1,61 @@
...
+        if (signal != null) {
+            value = signal;
+            reset();
+        }
+    }
+
+    /**
+     * This method reads the cached value.
+     *
+     * @return returns the cached value
```
aidnem: Please decide whether you want this method to destroy the cached value, and, if so, make sure you clearly document in this comment that it can 1) return null and 2) destroy the current value of the cache, making "reset" no longer work properly.
My personal preference would be that it doesn't actually replace the value with null, but rather returns null and stores the last value anyway.

aidnem (changes requested):
I didn't review the tests yet since I had some logic questions about the functionality of the main code.

math/src/main/java/coppercore/math/CachedDataTime.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    /**
+     * This method is used to write a new signal value(A new cache).
+     *
+     * @param signal the signal that is being sent.
+     */
+    public void write(Type signal) {
+        if (signal != null) {
+            value = signal;
```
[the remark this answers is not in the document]
godmar: CoPilot is right on this one.

math/src/test/java/coppercore/math/test/CachedDataReadTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+import coppercore.math.CachedDataRead;
+import edu.wpi.first.units.measure.Distance;
+import java.util.List;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class CachedDataReadTest {
+    @BeforeEach
```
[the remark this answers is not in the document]
godmar: CoPilot is correct here as well.

math/src/test/java/coppercore/math/test/CachedDataTimeTest.java
```diff
@@ -0,0 +1,118 @@
...
+    public void readTestTrue2() {
+        CachedDataTime data = new CachedDataTime(3);
+        data.write(-100);
+        setMockTimeSeconds(2.0);
+        Assertions.assertEquals(data.read(), -100);
+    }
+
+    @Test
+    public void readTestNull() {
+        CachedDataTime data = new CachedDataTime(1);
```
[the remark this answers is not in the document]
aidnem: Please use `Integer` here and not Object.

godmar (approved):
I refactored the shared parts into one class.  I think this is good enough and we could move on.

This cache basically doesn't distinguish between expired values and `null` values.

math/src/test/java/coppercore/math/test/CachedDataTimeTest.java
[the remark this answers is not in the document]
aidnem: I'm actually not too worried about this since I'd rather not delay this any further.

PR #136, opened 2025-03-15
Error when syncing primitive types with JSONSync

avidraccoon:


parameter_tools/src/main/java/coppercore/parameter_tools/json/JSONNamingStrategy.java
```diff
@@ -22,6 +24,32 @@ public JSONNamingStrategy(FieldNamingPolicy policy) {
...
         JSONName annotation = field.getAnnotation(JSONName.class);
+        Class<?> type = field.getType();
+        if (config.primitiveChecking()) {
+
+            if (type.isPrimitive()
+                    && !(Measure.class.isAssignableFrom(field.getDeclaringClass())
+                            && field.getDeclaringClass().isRecord())) {
+                if (config.primitiveCheckPrintAlert()) {
+                    Thread thread = new Thread(new JSONPrimitiveErrorAlert() {});
+                    thread.start();
```
aidnem, line 35: I feel that this thread alert is unnecessary, what if this feature just generated the same exception as primitive check crash but then printed its stack trace instead of throwing the exception?

aidnem (changes requested):
Just had 1 question

parameter_tools/src/main/java/coppercore/parameter_tools/json/JSONNamingStrategy.java
avidraccoon: Yes I will remove it

jkleiber:
@avidraccoon please don't tag the whole team on PR reviews in the future. Instead, I'd recommend reaching out to a team member to ask for a review and if they agree then tag them

avidraccoon:
@jkleiber  @aidnem  Should it have an annotation to ignore the primitive?

parameter_tools/src/main/java/coppercore/parameter_tools/json/JSONTypeAdapterFactory.java
```diff
@@ -14,10 +14,17 @@
 
 public class JSONTypeAdapterFactory implements TypeAdapterFactory {
 
+    private JSONSyncConfig config = new JSONSyncConfigBuilder().build();
```
godmar, line 17: Is this value overwritten in the constructor?

aidnem (dismissed):
Just a few comments

godmar (approved):
approve fix for Per unit failure.

PR #138, opened 2025-03-30
Better coprocessor connected

j0shuaS:
adds a check for all cameras and a check for specific camera


j0shuaS:
it will need to be tested irl

vision/src/main/java/coppercore/vision/VisionLocalizer.java
```diff
@@ -88,12 +88,25 @@ public boolean hasMultitagResult() {
...
+        }
+        return true;
+    }
+
+    /**
+     * @param camera
+     * @return specific camera input is connected
+     */
+    public boolean coprocessorConnected(int camera) {
+        return inputs[camera].connected;
```
avidraccoon, line 109: Add bounds check

avidraccoon (changes requested):

vision/src/main/java/coppercore/vision/VisionLocalizer.java
```diff
@@ -88,13 +91,31 @@ public boolean hasMultitagResult() {
...
+            }
+        }
+        return true;
+    }
+
+    /**
+     * @param camera
+     * @return specific camera input is connected
+     */
+    public boolean coprocessorConnected(int camera) {
```
aidnem: Could you rename this to cameraConnected? It feels more apt considering it's checking a single camera.

aidnem (dismissed):

vision/src/main/java/coppercore/vision/VisionLocalizer.java
```diff
@@ -86,11 +86,28 @@ public boolean hasMultitagResult() {
...
+        return true;
+    }
+
+    /**
+     * @param cameraIndex
+     * @return specific camera input is connected
+     */
+    public boolean cameraConnected(int cameraIndex) {
+        if (cameraIndex < 0 || cameraIndex >= inputs.length) {
+            throw new IllegalArgumentException("Camera index out of bounds" + cameraIndex);
```
aidnem: Can we return false here and just print this message to sys.err instead? This way we won't crash code if we poorly misconfigure cameras at competition.

gradle.properties
```diff
@@ -1,2 +1 @@
-version=2026.1.0
-systemProp.file.encoding=utf-8
\ No newline at end of file
+version=2026.1.1
```
aidnem: It looks like the file encoding got deleted here, could you add it back?

aidnem (approved):

PR #139, opened 2025-03-31
Move command from 2025-robot-code to coppercore

avidraccoon:


avidraccoon:
@aidnem @godmar I dont think we even need this do we?

PR #146, opened 2025-04-17
145 make jsonsync more powerful

avidraccoon:
Add support for java polymorphism (Basicly determine type from a json value)
Add ability to add custom type adapters outside of copper core. (A feature I don't expect to be used but could be useful)


jkleiber:
@avidraccoon - could you update the PR description to give some high level overview of what this change is before I review this?

avidraccoon:
@jkleiber is the new description good?

parameter_tools/src/main/java/coppercore/parameter_tools/json/JSONSyncConfigBuilder.java
```diff
@@ -68,12 +78,39 @@ public JSONSyncConfigBuilder setLongSerializationPolicy(
...
+        typeAdapters.add(new Pair<>(clazz, adapter));
+        return this;
+    }
+
+    public <T> JSONSyncConfigBuilder addJsonTypeAdapter(Class<T> clazz, TypeAdapter<T> adapter) {
+        typeAdapters.add(new Pair<>(clazz, adapter));
+        return this;
+    }
+
+    public <T> JSONSyncConfigBuilder settupPolymorphAdapter(Class<T> clazz) {
```
jkleiber, line 97: `setUpPolymorphAdapter` -> Spelling pls

jkleiber (commented):
Thanks for the updated description, that made it much easier to review. Also, could you add a unit test for the custom type adapter case? I think you added one for the polymorphism case only

avidraccoon:
> Thanks for the updated description, that made it much easier to review. Also, could you add a unit test for the custom type adapter case? I think you added one for the polymorphism case only

Both the custom adapter and polymorphism use the same interface behind the scene only reason there is separate methods is for type safety and not having to guess the type it wants.

jkleiber (approved):

PR #150, opened 2025-04-28
Normalize deadband values

linglejack06:
- add calculation (value - deadband) / (1 - deadband)
- this should help controller be less jerky


avidraccoon (dismissed):

wpilib_interface/src/main/java/coppercore/wpilib_interface/DriveWithJoysticks.java
```diff
@@ -103,17 +104,16 @@ public void execute() {
...
-     *  @return Translation2d with directions of velocity
+     * @return Translation2d with directions of velocity
      */
     public Translation2d getLinearVelocity(double x, double y) {
         double[] deadbands = Deadband.twoAxisDeadband(x, y, joystickDeadband);
 
-        double xDeadband = deadbands[0];
-        double yDeadband = deadbands[1];
+        double xDeadband = (deadbands[0] - joystickDeadband) / (1 - joystickDeadband);
+        double yDeadband = (deadbands[1] - joystickDeadband) / (1 - joystickDeadband);
```
aidnem, line 116: @linglejack06 will this work correctly for negative values? This pushes all measurements in the negative directions, so I think it would actually make you drive *faster* when driving left/down

aidnem (commented):
Just 1 question

wpilib_interface/src/main/java/coppercore/wpilib_interface/DriveWithJoysticks.java
aidnem: As evidence, here's a graph of the output of this function with joystick input as the X axis:

image

It looks like it works correctly for positive but doubles the error for negative.

aidnem: I think this line should work, we negate the subtraction if our input is negative:

```java
        double xDeadband = (deadbands[0] - Math.signum(deadbands[0]) * joystickDeadband) / (1 - joystickDeadband);
```

aidnem:
@godmar this a little lower priority, but would you mind reviewing it if you have time? It's a small improvement but it could have a large impact on fine robot control if next year's game requires fine-tuned movements. I did this now for 2 reasons:
    1) To finally get another PR out of "waiting for review" (especially since Jack graduated)
    2) To get a little break from the thousands of lines of code I've written for #164

wpilib_interface/src/main/java/coppercore/wpilib_interface/DriveWithJoysticks.java
```diff
@@ -116,6 +123,7 @@ public void execute() {
         Translation2d linearSpeeds = getLinearVelocity(-leftJoystickX, -leftJoystickY);
 
         double omega = Deadband.oneAxisDeadband(-rightJoystickX, joystickDeadband);
+        omega = (omega - Math.signum(omega) * joystickDeadband) / (1 - joystickDeadband);
```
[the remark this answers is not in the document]
godmar: Since the plan is to remove this functionality from coppercore, resolving this.

wpilib_interface/src/test/java/DriveWithJoysticksTests.java
```diff
@@ -0,0 +1,191 @@
...
+                "DriveWithJoysticks should always be requesting field centric speeds.");
+    }
+
+    private static double SPEED_EPSILON = 1e-6;
+
+    /**
+     * Assert that each field of a given set of ChassisSpeeds is within SPEED_EPSILON of an expected
+     * value.
+     *
+     * <p>Using AssertEquals seems to try pointer comparison and fail, so this method is required
```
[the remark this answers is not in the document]
godmar: Weird, but you're right, ChassisSpeed's `.equals` compares doubles with `==`.

aidnem:
> @aidnem do you agree?

@godmar Yes. Thanks for writing this up so the status is clear at a glance.

Do we want to add another step to make the exponent configurable (instead of forcing squaring)?

godmar:
> @godmar yes, given the configurable exponent in WPILib, do we want to use this PR to make DriveWithJoysticks take a configurable exponent rather than blindly squaring all inputs?

Sure, the width of the deadband could be passed then, too, somehow. Could be a factory that returns a joystick drive command or similar.

thetaback:
I checked out a fresh copy of coppercore and this branch, and I am still seeing the JVM crash when I do `./gradlew test`.
```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  EXCEPTION_ACCESS_VIOLATION (0xc0000005) at pc=0x00007ffba8b72f58, pid=24344, tid=25920
#
# JRE version: Java(TM) SE Runtime Environment (17.0.12+8) (build 17.0.12+8-LTS-286)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (17.0.12+8-LTS-286, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, windows-amd64)
# Problematic frame:
# C  [msvcp140.dll+0x12f58]
#
# No core dump will be written. Minidumps are not enabled by default on client versions of Windows
#
# If you would like to submit a bug report, please visit:
#   https://bugreport.java.com/bugreport/crash.jsp
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#

---------------  S U M M A R Y ------------

Command Line: -Djava.library.path=C:\Users\theta\robotics\coppercore\controls\build\jni\release -Djunit.jupiter.extensions.autodetection.enabled=true -Dorg.gradle.internal.worker.tmpdir=C:\Users\theta\robotics\coppercore\controls\build\tmp\test\work -Dorg.gradle.native=false -Xmx512m -Dfile.encoding=windows-1252 -Duser.country=US -Duser.language=en -Duser.variant -ea worker.org.gradle.process.internal.worker.GradleWorkerMain 'Gradle Test Executor 4'

Host: AMD Ryzen AI 7 PRO 350 w/ Radeon 860M          , 16 cores, 55G,  Windows 11 , 64 bit Build 26100 (10.0.26100.7309)
Time: Sun Jan 11 11:33:32 2026 Eastern Standard Time elapsed time: 0.752749 seconds (0d 0h 0m 0s)

---------------  T H R E A D  ---------------

Current thread (0x00000163ab1da350):  JavaThread "Test worker" [_thread_in_native, id=25920, stack(0x0000002b05500000,0x0000002b05600000)]

Stack: [0x0000002b05500000,0x0000002b05600000],  sp=0x0000002b055f9c90,  free space=999k
Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
C  [msvcp140.dll+0x12f58]
C  [wpiHal.dll+0x1a2c]

Java frames: (J=compiled Java code, j=interpreted, Vv=VM code)
```

godmar:
The other Spark deprecations are annoying. Looks like their 2026 API is more complex than before, as if they're trying to emulate CTRE's...  For instance, [`ResetMode.kResetSafeParameters`](https://codedocs.revrobotics.com/java/com/revrobotics/spark/sparkbase.resetmode) just says "deprecated" without saying what replaces it.

godmar:
> @godmar I am developing on windows 11 and sometimes macOS (on m1). I'm not seeing any segfaults or anything.
> 

The problem, I believe, is that it's loading the `msvcp140.dll` that's distributed with Java 17 as opposed to the one that's bundled with wpi.  Perhaps some gradle setting (?). (There are about 40 copies of `msvcp140.dll` on this new (!) machine spanning an estimated 5-8 different versions; talk about [DLL Hell](https://en.wikipedia.org/wiki/DLL_hell)).
She installed only JDK 17 on that machine since that's the version RoboRIO uses this year.

thetaback:
I was able to pick up the correct WPILib JDK by opening a git bash terminal in the WPILib VS Code and run `./gradlew test`. I think this might be better than to set the default JDK to the WPILib as I would like to keep the default settings unchanged for my own projects.

aidnem:
@thetaback could you try adding a custom value for JAVA_HOME in your vscode settings for that specific project?
Something like this:
```json
"terminal.integrated.env.windows": {
  "JAVA_HOME": "C:\\Users\\Public\\wpilib\\2026\\jdk"
}
```
could be added to `.vscode/settings.json` and that might resolve the issue for you. If it does, feel free to make a PR to save future devs the struggle.

(I'm unsure of the backslashes and how they interact with JSON & vscode & Powershell, but hopefully you can figure out any issues that I may have caused with that).

No pressure if you're fine with just using Git Bash from VSCode, that works too.

PR #152, opened 2025-05-04
Add method to change max speeds of DriveWithJoysticks

aidnem:
- Our demo/child/comp mode speed settings needed a way to vary the speed and creating a new command every time upset me on a personal level, so I added a method to set the max speeds.

- Also rewrites most of the doc comments because I thought they would benefit from some rephrasing


jkleiber (approved):
changes seem reasonable, approved

PR #159, opened 2025-09-10
Update JSONConverter.java to support frequency unit

j0shuaS:


avidraccoon:
This is fixed in #121

PR #161, opened 2025-09-15
Updating state machine

avidraccoon:


avidraccoon:
Not needed anymore as we are rewriting statemachine in #168

PR #166, opened 2025-11-01
Add generalized MotorIO and MotorIOTalonFX implementation with mechanism config infrastructure

aidnem:
### Dependencies

- Adds Phoenix 6 as a vendor dependency to wpilib_interface

### Features

Adds:
- A base MotorIO interface with a MotorInputs class
- A base MotorIOTalonFX implementation that supports position and velocity control with no profile, trapezoidal/s-curve profiles, and motion magic expo profiles
- Config classes for configuring mechanisms, associating CAN IDs with their can bus names to prevent CAN errors, and vendor-agnostic motion profile configs in case we add support for other vendor's hardware in the future

### DIFF
--- a/math/build.gradle
+++ b/math/build.gradle
@@ -18,6 +18,9 @@ dependencies {
 	simulationRelease wpi.sim.enableRelease()
 }
 
+// injects the JNI extraction dir into java.library.path for tests.
+wpi.java.configureTestTasks(test)
+
 mavenPublishing {
 	pom {
 		description = 'FRC 401 supplemental math functions'
--- a/math/src/main/java/coppercore/math/CachedDataRead.java
+++ b/math/src/main/java/coppercore/math/CachedDataRead.java
@@ -1,0 +1,48 @@
+package coppercore.math;
+
+/**
+ * This class is used for temporarily storing data in a cache and determining whether or not it is
+ * stale using readcount-based expiration.
+ *
+ * @param <Type> the type of the value to cache
+ */
+public class CachedDataRead<Type> extends CachedDataValue<Type> {
+    private int readCount = 0;
+    private int maxReads;
+
+    /**
+     * This is a constructor for the readCount-based expiration of the Cached Data.
+     *
+     * @param maxReads the maximum amount of reads we can take before the cache is stale
+     */
+    public CachedDataRead(int maxReads) {
+        this.maxReads = maxReads;
+    }
+
+    /**
+     * This method increases the readCount and then returns the cached value unless it has expired.
+     *
+     * @return returns the cached value if not stale, else null.
+     */
+    public Type read() {
+        readCount++;
+        return super.read();
+    }
+
+    /**
+     * This method checks if the cache is stale(readcount). If the readCount is more than the
+     * maximum number of reads.
+     *
+     * @return true if the cache is stale, false if not
+     */
+    @Override
+    public boolean isStale() {
+        return readCount > maxReads;
+    }
+
+    /** This method resets the expiration tracking value (readCount) */
+    @Override
+    protected void reset() {
+        readCount = 0; // Reset the read count for read-based expiration
+    }
+}
--- a/math/src/main/java/coppercore/math/CachedDataTime.java
+++ b/math/src/main/java/coppercore/math/CachedDataTime.java
@@ -1,0 +1,47 @@
+package coppercore.math;
+
+import edu.wpi.first.math.MathSharedStore;
+
+/**
+ * This class is used for temporarily storing data in a cache and determining whether or not it is
+ * stale using time-based expiration.
+ *
+ * @param <Type> the type of the value to cache
+ */
+public class CachedDataTime<Type> extends CachedDataValue<Type> {
+
+    /** The last time the cache was written in milliseconds */
+    private double lastUpdateTimestampMs = 0.0;
+
+    /** The amount of time it takes a cache to be stale in milliseconds */
+    private double staleTimeMs = -1.0;
+
+    /**
+     * This is a constructor for the Time-based expiration of the Cached Data.
+     *
+     * @param staleTimeSeconds the amount of time it takes a cache to be stale in seconds
+     */
+    public CachedDataTime(double staleTimeSeconds) {
+        this.staleTimeMs = staleTimeSeconds * 1000; // Convert seconds to milliseconds
+        this.lastUpdateTimestampMs = MathSharedStore.getTimestamp() * 1000;
+    }
+
+    /**
+     * This method checks if the cache is stale based on time.
+     *
+     * @return true if the cache is stale, false if not
+     */
+    @Override
+    public boolean isStale() {
+        double currentTime = MathSharedStore.getTimestamp() * 1000;
+        return currentTime - lastUpdateTimestampMs >= staleTimeMs;
+    }
+
+    /** This method resets the expiration tracking value for time based expiration. */
+    @Override
+    protected void reset() {
+        lastUpdateTimestampMs =
+                MathSharedStore.getTimestamp()
+                        * 1000; // Reset the last update time for time-based expiration
+    }
+}
--- a/math/src/main/java/coppercore/math/CachedDataValue.java
+++ b/math/src/main/java/coppercore/math/CachedDataValue.java
@@ -1,0 +1,44 @@
+package coppercore.math;
+
+/**
+ * This class is used for holding a cached value; subclasses implement specific expiration policies.
+ *
+ * @param <Type> the type of the value to cache
+ */
+public abstract class CachedDataValue<Type> {
+    private Type value = null;
+
+    /**
+     * Write a new data value (a cache entry) and reset expiration policy.
+     *
+     * @param data the new value to write to the cache.
+     */
+    public void write(Type data) {
+        if (data != null) {
+            value = data;
+            reset();
+        }
+    }
+
+    /**
+     * Returns cached value.
+     *
+     * @return returns the cached value or null if the value is stale.
+     */
+    public Type read() {
+        if (isStale()) {
+            return null;
+        }
+        return value;
+    }
+
+    /**
+     * This method checks if the cache is stale.
+     *
+     * @return true if the cache is stale, false if not
+     */
+    protected abstract boolean isStale();
+
+    /** Reset expiration policy for this value */
+    protected abstract void reset();
+}
--- a/math/src/test/java/coppercore/math/test/CachedDataReadTest.java
+++ b/math/src/test/java/coppercore/math/test/CachedDataReadTest.java
@@ -1,0 +1,88 @@
+package coppercore.math.test;
+
+import static edu.wpi.first.units.Units.Inches;
+import static edu.wpi.first.units.Units.Millimeters;
+
+import coppercore.math.CachedDataRead;
+import edu.wpi.first.units.measure.Distance;
+import java.util.List;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class CachedDataReadTest {
+
+    @SuppressWarnings("unused")
+    @Test
+    public void isStaleTest1() {
+        CachedDataRead<Integer> data = new CachedDataRead<>(5);
+        for (int i = 0; i < 5; i++) {
+            Assertions.assertFalse(data.isStale());
+            data.read();
+        }
+        data.read();
+        Assertions.assertTrue(data.isStale());
+
+        // Second Test
+        data = new CachedDataRead<>(20);
+        // 20 reads
+        for (int i = 0; i < 20; i++) {
+            Assertions.assertFalse(data.isStale());
+            data.read();
+        }
+        data.read();
+        Assertions.assertTrue(data.isStale());
+        // Ensure stale remains true
+        for (int i = 0; i < 10; i++) {
+            data.read();
+            Assertions.assertTrue(data.isStale());
+        }
+    }
+
+    @SuppressWarnings({"rawtypes"})
+    @Test
+    public void testReadReturnsLastWrittenValueForVariousDataTypes() {
+        CachedDataRead<String> stringData = new CachedDataRead<>(10);
+        stringData.write("real");
+        Assertions.assertEquals(stringData.read(), "real");
+        stringData.write("fake");
+        Assertions.assertEquals(stringData.read(), "fake");
+        stringData.write(null);
+        Assertions.assertEquals(stringData.read(), "fake");
+        CachedDataRead<Integer> integerData = new CachedDataRead<>(10);
+        integerData.write(100);
+        Assertions.assertEquals(integerData.read(), 100);
+        CachedDataRead<Boolean> booleanData = new CachedDataRead<>(10);
+        booleanData.write(true);
+        Assertions.assertEquals(booleanData.read(), true);
+        booleanData.write(false);
+        Assertions.assertEquals(booleanData.read(), false);
+        CachedDataRead<Distance> distanceData = new CachedDataRead<>(10);
+        distanceData.write(Millimeters.of(5));
+        Assertions.assertEquals(distanceData.read(), Millimeters.of(5));
+        distanceData.write(Inches.of(5));
+        Assertions.assertEquals(distanceData.read(), Inches.of(5));
+        CachedDataRead<List> listData = new CachedDataRead<>(10);
+        listData.write(List.of(1, 0, 0, 0, 1, -1));
+        Assertions.assertEquals(listData.read(), List.of(1, 0, 0, 0, 1, -1));
+        integerData.write(-100);
+        Assertions.assertEquals(integerData.read(), -100);
+        integerData.write(0);
+        Assertions.assertEquals(integerData.read(), 0);
+    }
+
+    @Test
+    public void readTestNull() {
+        CachedDataRead<Integer> integerData = new CachedDataRead<>(10);
+        CachedDataRead<String> stringData = new CachedDataRead<>(10);
+        integerData.write(100);
+        stringData.write(null);
+        for (int i = 0; i < 10; i++) {
+            Assertions.assertEquals(integerData.read(), 100);
+        }
+        for (int i = 0; i < 10; i++) {
+            Assertions.assertEquals(stringData.read(), null);
+        }
+        Assertions.assertEquals(integerData.read(), null);
+        Assertions.assertEquals(stringData.read(), null);
+    }
+}
--- a/math/src/test/java/coppercore/math/test/CachedDataTimeTest.java
+++ b/math/src/test/java/coppercore/math/test/CachedDataTimeTest.java
@@ -1,0 +1,115 @@
+package coppercore.math.test;
+
+import static edu.wpi.first.units.Units.Inches;
+import static edu.wpi.first.units.Units.Microseconds;
+import static edu.wpi.first.units.Units.Millimeters;
+import static edu.wpi.first.units.Units.Seconds;
+
+import coppercore.math.CachedDataTime;
+import edu.wpi.first.units.measure.Distance;
+import edu.wpi.first.util.WPIUtilJNI;
+import java.util.List;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class CachedDataTimeTest {
+
+    @BeforeEach
+    @SuppressWarnings("unused")
+    void setUpMockTime() {
+        WPIUtilJNI.enableMockTime();
+        setMockTimeSeconds(0.0);
+    }
+
+    /**
+     * Set WPIUtilJni's mock time to timeSeconds
+     *
+     * <p>This exists as a wrapper because WPIUtilJNI.setMockTime expects the time as a long in
+     * microseconds
+     *
+     * @param timeSeconds the time to set, in seconds
+     */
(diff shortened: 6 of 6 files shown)


wpilib_interface/src/main/java/coppercore/wpilib_interface/subsystems/PositionMechanism.java
```diff
@@ -0,0 +1,49 @@
...
+        this.inputs = inputs;
+        this.io = io;
+
+        this.followerIos = followerIos;
+
+        // Config all followers to follow leader motor
+        for (T follower : followerIos) {
+            follower.follow(config.leadMotorId, false);
+        }
+    }
```
avidraccoon: Maybe make sure the config follower ids and follower IOs match length and the ids match.

wpilib_interface/src/main/java/coppercore/wpilib_interface/subsystems/configs/MechanismConfig.java
```diff
@@ -0,0 +1,33 @@
...
+     *
+     * @param name Mechanism name, used for logging
+     * @param canbus Name of the canbus to use for all devices (usually "rio" or "canivore")
+     * @param leadMotorId CAN ID of the lead motor
+     * @param followerMotorIds CAN IDs of the follower motors. Leave empty for no followers.
+     */
+    public MechanismConfig(String name, String canbus, int leadMotorId, int[] followerMotorIds) {
+        this.name = name;
+        this.canbus = canbus;
+        this.leadMotorId = leadMotorId;
```
godmar: I actually liked Team 254's approach of abstracting the combination of can bus and can bus id into a "CanDeviceId" class.  Should we consider this?

wpilib_interface/src/main/java/coppercore/wpilib_interface/subsystems/configs/MechanismConfig.java
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Create a new base mechanism config.
+     *
+     * @param name Mechanism name, used for logging
+     * @param leadMotorId CAN ID of the lead motor
+     * @param followerMotorConfigs Array of configs for follower IDs and inverts. Leave empty for no
+     *     followers.
+     */
+    public MechanismConfig(
```
godmar: Should we try a bit of the builder pattern here, by allowing for a ".addFollower(...)" method rather than requiring the user construct an array of `MechanismFollowerMotorConfig`.  In fact, the entire `MechanismConfig` might benefit from the builder pattern.

On the other hand, `MechanismConfig` is likely immutable, and it might make sense to check its completeness upon construction. Towards that end, perhaps we could add `Objects.requireNonNull()` as appropriate.

wpilib_interface/src/main/java/coppercore/wpilib_interface/subsystems/motors/profile/ImmutableMotionProfileConfig.java
```diff
@@ -0,0 +1,133 @@
...
+    /**
+     * A generic, immutable motion profile configuration
+     *
+     * <p>For a mutable config, see {@link MotionProfileConfig#mutable}
+     *
+     * <p>Leaving any value at 0 signifies that it is uncapped
+     *
+     * <p>This class is almost identical to a Phoenix-6 MotionMagicConfigs object except that a
+     * coppercore MotionProfileConfig stores all values that can reasonably have units attached with
+     * units, whereas MotionMagicExpoConfigs use doubles.
```
godmar: Why then are `expoKv` and `expoKa` stored as doubles here instead of `Voltage` and `Current`?

wpilib_interface/src/main/java/coppercore/wpilib_interface/subsystems/motors/profile/ImmutableMotionProfileConfig.java
```diff
@@ -0,0 +1,133 @@
...
+        this.expoKv = other.getExpoKv();
+        this.expoKa = other.getExpoKa();
+    }
+
+    /**
+     * Converts this motion profile config into a motion magic config
+     *
+     * @return a MotionMagicConfigs object with fields from this configuration
+     */
+    public MotionMagicConfigs asMotionMagicConfigs() {
```
godmar: This is specific to the TalonFX motors, isn't it?  Add a comment to that effect.

wpilib_interface/src/main/java/coppercore/wpilib_interface/subsystems/motors/profile/MutableMotionProfileConfig.java
```diff
@@ -0,0 +1,245 @@
...
+     * <pre>{@code
+     * MutableMotionProfileConfig.zeros()
+     *     .withExpoKv(0.5)
+     *     .withExpoKa(0.2)
+     *     .freeze();
+     * }</pre>
+     *
+     * @return The new copy
+     */
+    // This method can't be implemented here because the Mutable config has to copy its measures
```
godmar: What does this comment mean? It seems the method is implemented here.

wpilib_interface/src/main/java/coppercore/wpilib_interface/subsystems/motors/MotorIOSparkMax.java
```diff
@@ -0,0 +1,126 @@
...
+        connected &=
+                SparkUtil.ifOk(
+                        sparkMax,
+                        () -> sparkMax.getAlternateEncoder().getPosition(),
+                        (value) -> inputs.position.mut_replace(value, Rotations));
+        connected &=
+                SparkUtil.ifOk(
+                        sparkMax,
+                        () -> sparkMax.getAlternateEncoder().getPosition(),
+                        (value) -> inputs.position.mut_replace(value, Rotations));
```
godmar: is this a copy and paste mistake?  Reads identical.

wpilib_interface/src/main/java/coppercore/wpilib_interface/subsystems/motors/MotorIOSparkMax.java
```diff
@@ -0,0 +1,126 @@
...
+                        sparkMax,
+                        sparkMax::getOutputCurrent,
+                        (value) -> inputs.statorCurrentAmps = value);
+        connected &=
+                SparkUtil.ifOk(
+                        sparkMax,
+                        () -> sparkMax.getAbsoluteEncoder().getPosition(),
+                        (value) -> inputs.rawRotorPosition.mut_replace(value, Rotations));
+
+        inputs.connected = connected;
```
godmar: Does this mean a SparkMax has to have *all* of these to be considered "connected?"

godmar (changes requested):
I think this is overall great.

Please check whether we need mutable profiles. If we can keep it more simple without them I would go for that.

We'll need a showcase/demo example of how to use the new MotorIOs.
Also, can some of this unit tested?

wpilib_interface/src/main/java/coppercore/wpilib_interface/subsystems/motors/MotorIOSparkMax.java
aidnem: As written, yes. I'm attempting to mimic the behavior of the TalonFX IO, where if refreshing any status signal fails to refresh then that indicates a connectivity issue. However, you raise a good point: what if we want to use motors without attaching an external encoder? Perhaps I could give the config a way to specify whether or not an external encoder is required? 

It's cases like these where I wish i could create something like a Rust enum for this:

```java
// MechanismConfig.java

    enum RemoteEncoder {
        None,
        CANRemoteEncoder(CANDeviceID id)
        SparkExternalEncoder
    }

    public final RemoteEncoder remoteEncoder;
```

I could represent this with a sealed interface that permits classes with the right fields, or just create a class to represent the remote encoder config that uses a factory to only allow you to instantiate correctly-formed objects: a None/ExternalEncoder config wouldn't need a CAN ID, but a CANRemoteEncoder would.

Do you have a preference for how to represent this?

wpilib_interface/src/main/java/coppercore/wpilib_interface/subsystems/motors/MotorIO.java
```diff
@@ -0,0 +1,185 @@
...
+     * the motor.
+     *
+     * @param profileConfig The motion profile configuration to use.
+     */
+    public void setProfileConstraints(MotionProfileConfig profileConfig);
+
+    /**
+     * Set the closed-loop control gains used by the motor.
+     *
+     * <p>This method should be assumed to be blocking, and may take significant time to finish. If
```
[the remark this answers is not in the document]
PChild: If this is coming from [Applying Configs](https://v6.docs.ctr-electronics.com/en/latest/docs/api-reference/api-usage/configuration.html#applying-configs) for Talon FX they do include a warning about it and applying configs to 16 motors at the start of auto did cause a meaningful (maybe 500 ms) delay for us last year.

wpilib_interface/src/main/java/coppercore/wpilib_interface/subsystems/motors/MotorIOTalonFX.java
```diff
@@ -0,0 +1,285 @@
...
+                                .withKV(kV)
+                                .withKA(kA)
+                                .withGravityType(
+                                        config.gravityFeedforwardType
+                                                .toPhoenix6GravityTypeValue()));
+    }
+
+    @Override
+    public void setBrakeMode(boolean shouldBrake) {
+        talon.setNeutralMode(shouldBrake ? NeutralModeValue.Brake : NeutralModeValue.Coast);
```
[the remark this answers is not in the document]
godmar: I think it would slightly better but not strictly necessary. I recall wondering what a true vs a false brake mode was the first time I read it. Make a call and go with it.

wpilib_interface/src/main/java/coppercore/wpilib_interface/subsystems/motors/MotorIOTalonFX.java
```diff
@@ -0,0 +1,285 @@
...
+     * @param followerIndex An Optional containing either the index of the follower motor (what
+     *     position in config.followerIds this motor is) or None if this is the lead motor. If
+     *     followerIndex is not None, this IO will automatically follow the lead motor at the end of
+     *     its constructor.
+     * @param talonFXConfig A TalonFXConfiguration to apply to the motor. This config will not be
+     *     modified by this IO, so there's no need to copy it.
+     */
+    public MotorIOTalonFX(
+            MechanismConfig config,
+            Optional<Integer> followerIndex,
```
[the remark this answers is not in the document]
aidnem: I've added `.addFollower` to the builder. How would you envision that affecting this constructor? Do you think I should write some sort of factory to create these IOs for the user? In our robot code, the IOs are instantiated at runtime depending on what mode the robot is in, so I had planned for the user to have to create the IO as needed (or even extend it for their own needs).

wpilib_interface/src/main/java/coppercore/wpilib_interface/subsystems/motors/profile/MutableMotionProfileConfig.java
```diff
@@ -0,0 +1,245 @@
...
+ *       output required to maintain a certain velocity. Is in units of Output Unit / Rotations per
+ *       Second. As a result, when supply voltage is fixed, a higher profile kV results in a lower
+ *       profile velocity.
+ *   <li><b>expoKa</b> - The kA for MotionMagicExpo/exponential motion profile. Represents the
+ *       output required to apply a given acceleration. Is in units of Output Unit / (Rotations per
+ *       Second / Second). As a result, when supply voltage is fixed, a higher profile kA results in
+ *       a lower profile acceleration.
+ * </ul>
+ */
+public final class MutableMotionProfileConfig extends MotionProfileConfig {
```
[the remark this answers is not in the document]
aidnem: This would only be necessary in the case that you're needing to actually calculate the speeds at runtime, so there cannot be an immutable profile, unless you're creating it every periodic. If you were to create one every periodic, you might run into issues eventually because you would be creating new objects so often, so I thought I'd give the user the option to just use one config and then update it dynamically. If you think this is too much premature optimization/won't be a concern, I can remove it, since we'll likely not use it anyway.

wpilib_interface/src/main/java/coppercore/wpilib_interface/subsystems/motors/MotorIO.java
```diff
@@ -0,0 +1,185 @@
...
+            double expoKa);
+
+    /**
+     * Control the motor to a certain position setpoint using closed-loop control and a motion
+     * profile
+     *
+     * @param positionSetpoint The setpoint to control the motor to
+     * @param profileConfig
+     */
+    public void controlToPositionProfiled(
```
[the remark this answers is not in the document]
aidnem: If we get rid of the overload with individual arguments, we'll either have to keep MutableMotionProfileConfig or accept that people will have to create a MotionProfileConfig every periodic to do dynamic profiling (or just extend MotorIO with the features they need). In my opinion, we should probably pick one to keep, unless we're certain that the performance will never be a concern.

wpilib_interface/src/main/java/coppercore/wpilib_interface/subsystems/motors/profile/MutableMotionProfileConfig.java
aidnem: This was a copy paste error.

wpilib_interface/src/main/java/coppercore/wpilib_interface/subsystems/motors/profile/ImmutableMotionProfileConfig.java
aidnem: I discovered that we have predefined units for `VoltsPerRadianPerSecond` and `VoltsPerRadianPerSecondSquared`. I went ahead and added `VoltsPerRotatonPerSecond` and `VoltsPerRotationPerSecondSquared` into UnitUtils since I generally prefer rotations.

wpilib_interface/src/main/java/coppercore/wpilib_interface/subsystems/motors/MotorIOSparkMax.java
aidnem: I think I'll add an optional field to the config that tells it there is no external encoder present for a Spark device. TalonFX can run closed-loop with internal encoder but I don't believe SparkMax can since you need to plug in the motor encoder depending on the device.

According to Preston:
> A spark max uses the normal encoder port pins to connect to the hall effect sensor board inside a NEO and that's what gets used for position and velocity closed loop most of the time. Alternate encoder mode remaps some of the pins in the data port so that you can use a quadrature encoder there and not interfere with the pins the NEO uses. The spark should use the external encoder when configured in external encoder mode (this implies a brushed not brushless motor) and when in alternate encoder mode too.

PChild: I'm not following what you've written Aiden. Internal encoder == motor encoder. The Spark Max can do closed loop position and velocity with the encoder in the NEO.

aidnem: @PChild somehow in all of my searching I missed that the whole "motor encoder" existed in this ecosystem. I also missed that there was simply a `getEncoder` method on the SparkMax. I'm fixing that now to use the embedded encoder for everything that isn't absolute, and then make "requiring" the absolute encoder optional so that you can still get a connectivity alert if the absolute encoder disconnects when you are using it.

aidnem:
I've gotten to a point where tests pass locally, but fail on CI for some reason.

aidnem:
@godmar I believe I'm going to downsize my unit tests to test whether or not the motor is applying a reasonable output and whether the IOs are correctly getting the details, such as translating gear ratios and such from a dummy sim adapter. In my opinion, it's too much of a struggle to get a whole sim system passing tests considering the fact that, at that point, it's not really a "unit" anymore: if that test fails, you know that either the IO, the sim IO, the position adapter, or wpilib itself has an error, but there's hardly any indication where it is. Any objections?

godmar:
Could you explain this test. Is the simulated setup such that the motor is attached to the same mechanism as the encoder? If so, would driving the motor (via simulation) be expected to change the encoder readings as the mechanism is moving?
```java
        simAdapter.setMotorPosition(Radians.of(10.0));     // motor is being told it's at 10 rad
        simAdapter.setEncoderPosition(Radians.of(2.0));   // encoder is being told it's at 2 rad

        // Enabling!
        DriverStationSim.setEnabled(true);
        DriverStationSim.notifyNewData();

        assert DriverStation.isEnabled();   // I hope the test framework runs with -ea or else this statement won't be executed

        // Give it a couple cycles to let data propagate into the IOs
        leadMotor.controlToPositionUnprofiled(Radians.zero());  // this tells the motor to drive to 0 rad position
        loopForTime(1.0, loop);

        // should this then change the encoder position via a connected mechanism?  If so, wy would we expect it to still be at 2 rad?
        Assertions.assertEquals(
                2.0,
                simAdapter.getEncoderPosition().in(Radians),
                1e-1); // Give it a decent delta to account for slight oscillation
 ```

aidnem:
@godmar The simulated setup is that the encoder is connected to the motor where the motor is at a 5:1 reduction from the encoder (5 motor rotations = 1 encoder rotation). When moving one, it would be expected that the other moves the same amount multiplied by the ratio.

The actual position measurements that they start at are probably not super relevant because the motor cannot be expected to act as an absolute encoder and thus the encoder would only have a 1/5 chance of being at 0 when the motor was at 0 on power on. However, I'm keeping them all on a 1:1 ratio to make it easier to keep the movements on a 1:1 ratio, which would actually be relevant in the case of a FusedCANCoder setup.

godmar:
> @godmar The simulated setup is that the encoder is connected to the motor where the motor is at a 5:1 reduction from the encoder (5 motor rotations = 1 encoder rotation). When moving one, it would be expected that the other moves the same amount multiplied by the ratio.
> 
> The actual position measurements that they start at are probably not super relevant because the motor cannot be expected to act as an absolute encoder and thus the encoder would only have a 1/5 chance of being at 0 when the motor was at 0 on power on. However, I'm keeping them all on a 1:1 ratio to make it easier to keep the movements on a 1:1 ratio, which would actually be relevant in the case of a FusedCANCoder setup.

I was wondering why you assert that `simAdapter.getEncoderPosition()` is still 2.0 rad even though you're simulating motor movement for a second. But I suppose that is because nothing updates the encoder position via the DummySimAdapter. Is the purpose of testing it that it stays at 2?

How exactly do you assert that `leadMotor.controlToPositionUnprofiled(Radians.zero())` has the desired effect?
The only relevant assert I see is that the motor should report a negative voltage (which fails on OSX).

aidnem:
>         // should this then change the encoder position via a connected mechanism?  If so, wy would we expect it to still be at 2 rad?

We don't expect the mechanism to move because, when using a DummySimAdapter, the position is always a constant until we change it.

> But I suppose that is because nothing updates the encoder position via the DummySimAdapter. Is the purpose of testing it that it stays at 2?

Precisely.

> How exactly do you assert that leadMotor.controlToPositionUnprofiled(Radians.zero()) has the desired effect?
The only relevant assert I see is that the motor should report a negative voltage (which fails on OSX).

My hope was that, when commanding it to command to a position below its position, it would try to control at least in the right direction. That's why I'm just asserting a negative voltage. I didn't think I would compare the magnitude because, depending on lag/timing, the motion profile would have progressed a variable distance. I also just found sim performance overall to be very buggy in the context of a unit test, while it was fairly reliable in an actual robot project.

I'm not sure why it's failing on MacOS. The tests pass when I run them on my personal mac. I can look into that further as well. Right now I'm trying to get the SparkMax sim IOs to work in a real robot project, since right now they have odd behavior and don't really work properly at all.

aidnem:
@godmar I've added CI to upload test reports from all tests so that we can at least see the behavior in CI. Unfortunately I'm still completely baffled by why this test should be failing on MacOS but not on any other system. It seems like the sim is just never being updated on the MacOS CI machine, as the encoder position reported (2.0) is equal to the position we're setting (motor position 10.0 / 5.0 ratio = 2.0).

To access the artifacts, click on a CI run:
image

Then expand this section to download the ZIP from the link shown:

image

godmar:
I've been looking at this too.  

Generally, speaking, the idea that 
- one thread sleeps for 20ms
- and expects some kind of guarantee that another thread will execute for some "sufficient" amount of CPU time during this pause

is fundamentally a fishy/broken idea.   This is because the OS will, if a CPU/core is available, try to schedule threads that are ready to run anyhow - but there's no guarantee this will happen during the next 20ms. It depends on how many CPUs are available, how many ready threads are contending for them, etc. Also, there's no guarantee about the accuracy of Thread.sleep (which is what `Timer.delay` calls).

The [documentation](https://v6.docs.ctr-electronics.com/en/latest/docs/api-reference/wpilib-integration/unit-testing.html) mentions to call `Timer.delay(0.100)` once and then `Timer.delay(0.02)` after each command, plus `Unmanaged.feedEnable(100)` (you're calling 20).

godmar (approved):
LGTM - we should probably wrap this up so that we have a new release of coppercore before kickoff?

aidnem:
@godmar to follow up on our lingering concerns regarding SparkMax sim IOs and real-life unit test accuracy, I've created issues #180 and #181.

PR #172, opened 2025-11-19
Replace hardcoded CoreVisionConstants with dependency-injected VisionGainConstants

aidnem:
## Changes

- Rename CoreVisionConstants to VisionGainConstants
- Convert fields from static to instance fields; make VisionLocalizer take a VisionGainConstants object as an argument
- Add "builder" style methods to update the config more easily
- Add javadocs to all fields.

## Verification

Here's a recording of CHCMP quals 88 after being simulated with replay sim, where the gains passed to the VisionLocalizer were the default gains. As you can see, the replay outputs ghost (orange) and the real outputs ghost (blue) behave identically.

https://github.com/user-attachments/assets/89124fbe-39a6-4e40-9c2c-a06c055f06b8

Now, here's a recording where the linearStdDevFactor has been increased from 0.02 to 1.0, making the replay have much more stable/less jittery pose estimation. You can see the real output ghost (blue) jittering while the replay outputs ghost (orange) is now staying relatively still.

https://github.com/user-attachments/assets/7ff3af02-8af2-4ecd-aac9-11448bb1aa93

### DIFF
--- a/vision/src/main/java/coppercore/vision/VisionGainConstants.java
+++ b/vision/src/main/java/coppercore/vision/VisionGainConstants.java
@@ -2,12 +2,21 @@ package coppercore.vision;
 
 /** Vision constants for coppercore's pose rejection and standard deviation calculations. */
 public final class VisionGainConstants {
+    /**
+     * Maximum distance for a pose estimate to be accepted.
+     *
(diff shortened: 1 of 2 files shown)


godmar (approved):
LGTM

PR #186, opened 2026-01-13
185 update jsoncontrollers to work better and usework with newer json sync api

avidraccoon:


wpilib_interface/src/main/java/coppercore/wpilib_interface/controllers/Controller.java
```diff
@@ -188,26 +267,40 @@ public double getValue() {
...
+     subtypes = {
+        @JsonSubtype(clazz = Button.class, name = "button"),
+        @JsonSubtype(clazz = Axis.class, name = "axis"),
+        @JsonSubtype(clazz = POV.class, name = "pov"),
+     }
+    )
+    /**
+     * Controller interface class
+     * 
+     * These represents the interfaces that the code will interact with
```
godmar: I don't understand this. What do you mean by "interfaces"? What is an "interface" in this context?

It's more important to have a good explanation of the purpose of a class here than mechanical javadocs.

godmar (changes requested):
I'm having a hard time making sense of your `ControllerInterface` - that's probably just a naming issue.  Would `ControlElement` be an appropriate term? Or `ControllerComponent`?

PR #187, opened 2026-01-13
163 add afterjsonload annotation to call method after json values are loaded

avidraccoon:


parameter_tools/src/main/java/coppercore/parameter_tools/json/annotations/AfterJsonLoad.java
```diff
@@ -0,0 +1,11 @@
+package coppercore.parameter_tools.json.annotations;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/** Annotation to have custom name for a field. */
```
godmar: Please add a long javadoc comment describing the purpose of @AfterJsonLoad and explaining its use.

godmar (changes requested):
Mostly good, but we need to add documentation to the new annotation where it is defined.

Issue #190, opened 2026-01-14
document and improve way to publish local coppercore


godmar:
As we develop, we will have a local copy of coppercore checked out.  I've find it somewhat difficult in the past to verify that/whether the robot code is indeed using the desired version of coppercore.
I'd like to (a) document and (b) make this process more robust.

Idea:

- when publishing to maven local, gradle automatically generates a Java class that contains the version
- the robot code can load this class and print and/or verify this version
- this way, under local development, the version number can be increased and the increased version number must have come from the local maven repo, not the central one.
- add a README describing the steps.

PR #191, opened 2026-01-14
(#190) add coppercore metadata package for version info

godmar:
First draft for issue #190


godmar:
Yes, this is not tested on Windows. @thetaback could you test it on your machine as well?

We could also include git commit info, though I didn't include the code I had GPT write because I wasn't sure whether running git is possible on all platforms.

Issue #193, opened 2026-01-15
Change from TorqueCurrentFOC requests to Voltage requests with FOC enabled in TalonFX motor IO


aidnem:
Based on preliminary testing, it is likely easier to tune voltage systems than current systems, and so we should replace our TorqueCurrentFOC requests with FOC-enabled Voltage requests in MotorIOTalonFX. This will also hopefully ease debugging of sim, since ClosedLoopOutput and AppliedVolts should be equal to each other.

PR #194, opened 2026-01-15
(#193) Replace all requests in MotorIOTalonFX with Voltage equivalents

aidnem:
Replace all requests in MotorIOTalonFX with the voltage equivalent and write out withEnableFOC(true) on all of them. While FOC is enabled by default, I find that it makes it more explicit to state it in the code for those curious enough to read it (and I seriously doubt it's impactful to performance).

- Bump version to 2026.1.1

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,2 +1,2 @@
-version=2026.1.0
+version=2026.1.1
 systemProp.file.encoding=utf-8
--- a/wpilib_interface/src/main/java/coppercore/wpilib_interface/subsystems/motors/talonfx/MotorIOTalonFX.java
+++ b/wpilib_interface/src/main/java/coppercore/wpilib_interface/subsystems/motors/talonfx/MotorIOTalonFX.java
@@ -11,16 +11,16 @@ import com.ctre.phoenix6.StatusSignal;
 import com.ctre.phoenix6.configs.Slot0Configs;
 import com.ctre.phoenix6.configs.TalonFXConfiguration;
 import com.ctre.phoenix6.controls.CoastOut;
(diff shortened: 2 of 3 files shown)


godmar:
More research:

- a [thread on CD](https://www.chiefdelphi.com/t/how-do-you-do-motion-control/501447/27?u=godmar) on this topic. In it, Ben Hall (working for CTRE) [comments](https://www.chiefdelphi.com/t/how-do-you-do-motion-control/501447/30?u=godmar):
> With all that said, for unprofiled position control, Voltage may be easier to tune since you can sometimes skip kD. For velocity control, TorqueCurrentFOC is definitely easier to tune (it’s often just kS+kP). And for a profiled position system, TorqueCurrentFOC can be easier to tune since you don’t need a kV, and kA can be found independently.

I'm not sure how to read this. Would the turrent position control be unprofiled or profiled?  When would this be run, btw?
Would the turret already start turning while the robot is still driving towards the/a shooting position? (Ideally, yes).

From the software point of view, it appears that there's a choice - presumably that's why both voltage and torquecurrent controls exist. So our coppercore library should ideally support both and be configurable.

Given the time pressure, of course, I'm not sure what the best next step is.  Share your thoughts. Perhaps @PChild has experience with voltage vs torquecurrent motion from past seasons?

aidnem:
@godmar when I tuned the elevator last year, it went great. My only real difficulties have been the bizarre behavior of our wrist (a single jointed arm) last year in sim and the strangely changing kS on the turret prototype.
This leads me to believe that I might just be doing it wrong. If I have time tomorrow at shop, I'll try following CTRE's advice exactly and if it goes well maybe we can hold off on this change.

PR #198, opened 2026-01-19
197 fix publish failure due to metadata package generation

godmar:
GPT recommended this; but how do we test this - can we craft a release from a branch?

Yes. Did this.


godmar:
Done

PR #205, opened 2026-01-21
Slight Upgrades to State Machine

avidraccoon:
This pull request enhances the state machine API in the `coppercore.controls.state_machine` package by improving type safety and adding more flexible transition condition definitions. The most important changes are:

### Transition Condition Builder Enhancements

* Added overloads for the `when` and `andWhen` methods in the `State` class to accept `BooleanSupplier` conditions, making it easier to define transitions based on simple boolean logic without requiring a `World` parameter. [[1]](diffhunk://#diff-97cd91a111e6e6b6d7ed001934efb00dcede8012437dd45a9dbf36b62172e53aR278-R288) [[2]](diffhunk://#diff-97cd91a111e6e6b6d7ed001934efb00dcede8012437dd45a9dbf36b62172e53aR311-R321)

### Type Safety Improvement

* Updated the `registerState` method in `StateMachine` to use a generic return type, ensuring that the specific type of the registered state is preserved and returned, improving type safety for callers.

### Minor Improvements

* Added missing import for `BooleanSupplier` in `State.java` to support the new overloads.

This description was generated by copilot.

### DIFF
--- a/controls/src/main/java/coppercore/controls/state_machine/State.java
+++ b/controls/src/main/java/coppercore/controls/state_machine/State.java
@@ -169,6 +169,13 @@ public abstract class State<World> {
      * @param world The current world state
      */
     protected final void _periodic(StateMachine<World> stateMachine, World world) {
+        if (finished) {
+            System.err.println(
+                    "State machine bug: finished state: "
+                            + this
+                            + " did not transition out of a finished state, skipping its periodic");
+            return;
+        }
         periodic(stateMachine, world);
     }
 


avidraccoon:
@godmar I just realized that if a State has said that it is finished and it is not immediately exited the periodic method is still called is this something we are fine with? Because the state can check in it's periodic that it is currently not finished.

godmar:
> For the tests, should `externalCondition` var be an instance field of `class Robot { }`?

Your comment seems to imply you don't want it in world, but why?

avidraccoon:
> Your comment seems to imply you don't want it in world, but why?

It is to simulate having a method outside the statemachine world. Like if `DriverStation::isAutonomous`

godmar:
> Also not sure if warning is the right choice because what if the reason it has not transitioned yet is that some external condition need to be met for it to leave its state?

Then it shouldn't "finish" - it should use a `when(...)` transition for that condition (?)

Issue #203, opened 2026-01-20
Support changing camera poses


aidnem:
## Summary

If we put a camera on the turret, we'll need a way to support localization or relative tag estimation with a moving camera.

## Work

- [ ] Figure out a way to support variable camera poses. The way I've seen it be done is 6328's alpha bot code, where they use a TimeInterpolatableBuffer that tracks the turret positions in the past. The way we'd implement this library-side is to just ask for a function that converts a timestamp into a camera transform.
- [ ] Implement it
- [ ] Test it out on the test drivebase or comp bot, depending on which we get

godmar:
Pointer to [`TimeInterpolatableBuffer`](https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/math/interpolation/TimeInterpolatableBuffer.html).

aidnem:
Completed in #213

PR #213, opened 2026-01-30
(#203) Added updateInputs overload for changing camera poses

thetaback:


vision/src/main/java/coppercore/vision/VisionIOPhotonReal.java
```diff
@@ -27,10 +29,18 @@ public class VisionIOPhotonReal implements VisionIO {
     public VisionIOPhotonReal(String name, Transform3d robotToCamera) {
```
aidnem: Would you mind updating the javadoc here to reflect what robotToCamera is now?

godmar:
My suggestions for the order in which to do things would be:
- get this merged; release a new coppercore
- get https://github.com/team401/2026-Robot-Code/pull/25 merged
- branch off of main to integrate vision into 2026 code base (in parallel with https://github.com/team401/2026-Robot-Code/pull/38 and https://github.com/team401/2026-Robot-Code/pull/41)

aidnem (commented):
I had one comment/suggestion/question but overall this looks good.

vision/src/main/java/coppercore/vision/VisionIOPhotonSim.java
```diff
@@ -3,49 +3,76 @@
...
+     * @see VisionIO#updateInputs(coppercore.vision.VisionIO.VisionIOInputs, DoubleFunction)
+     */
     @Override
-    public void updateInputs(VisionIOInputs inputs) {
+    public void updateInputs(
+            VisionIOInputs inputs, DoubleFunction<Optional<Transform3d>> robotToCameraAt) {
+        // If the camera doesn't already exist, it should be created at the given transform. If it
+        // already existed, it should be adjusted to the new transform, which may be the same as
+        // before if the camera is stationary.
+        if (cameraSim == null) {
```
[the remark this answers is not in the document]
aidnem: Ok, I see how the api would be cleaner this way.

> That said, the way the code is right now is that in the case of a constant transform, it would fill a TimeInterpolatableBuffer with identical Pose3d objects, then hopefully interpolate between them correctly.

Wouldn't we be able to just create a DoubleFunction that returned the transform (a constant) on the code side? Then we wouldn't even need a buffer for fixed cameras. We could even give the camera config an overloaded constructor so that we would just pass it a Transform3D.

We could make the camera IOs internal to coppercore, but I don't think I understand what problem that would solve. I *think* just overloading the camera config constructor should be enough?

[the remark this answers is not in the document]
aidnem: Oh, I believe I misunderstood you. Are you referring to the [implementation](https://javadocs.photonvision.org/release/src-html/org/photonvision/simulation/VisionSystemSim.html#line-191:~:text=191%20%20%20%20public%20boolean-,adjustCamera,-(PhotonCameraSim%20cameraSim%2C%20Transform3d)) of adjustCamera, which would be adding values needlessly to this TimeInterpolatableBuffer? Perhaps having a `fixed` flag which stops us from calling adjustCamera and then using factory methods to generate the camera configs would be the correct move after all. E.g.

```java
new VisionLocalizer(
   ...
   CameraConfig.fixed(new VisionPhotonSim(...), cameraTransform)
```

```java
new VisionLocalizer(
   ...
   CameraConfig.moving(new VisionPhotonSim(...), cameraTransformFunc)
```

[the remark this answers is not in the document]
thetaback: I used your suggestion of factory methods for creating the CameraConfigs in commit b34bd53. I also added a transform initializer method to VisionIOPhotonSim to add the camera without the awkward logic in updateInputs, and I added a cameraType field to avoid the use of adjustCamera for fixed cameras.

vision/src/main/java/coppercore/vision/VisionIOPhotonSim.java
```diff
@@ -3,49 +3,81 @@
...
+        }
+        visionSim.update(poseSupplier.get());
+        super.updateInputs(inputs, robotToCameraAt);
+    }
 
+    /**
+     * Creates a camera with the given initial transform. This should be called only once when the
+     * VisionIOPhotonSim is created. This is called for both mobile and stationary cameras.
+     *
+     * @param robotToCamera the initialize transform of the robot to the camera
```
godmar: typo: "the initial..."

vision/src/main/java/coppercore/vision/VisionIOPhotonReal.java
```diff
@@ -40,8 +38,12 @@ public void setAprilTagLayout(AprilTagFieldLayout tagLayout) {
         aprilTagLayout = tagLayout;
     }
 
+    /**
+     * @see VisionIO#updateInputs(coppercore.vision.VisionIO.VisionIOInputs, DoubleFunction)
+     */
```
aidnem: If you just leave no javadoc here, it'll inherit the javadoc from the original vision IO and then people won't have to click a link to read it.

vision/src/main/java/coppercore/vision/VisionLocalizer.java
```diff
@@ -165,8 +241,10 @@ public DistanceToTag getDistanceErrorToTag(
     /** Periodically updates the camera data and processes new measurements. */
     @Override
     public void periodic() {
-        for (int i = 0; i < io.length; i++) {
-            io[i].updateInputs(inputs[i]);
+        for (int i = 0; i < cameras.length; i++) {
+            final int _i = i;
+            cameras[i].io.updateInputs(
+                    inputs[i], (time) -> cameras[_i].robotToCameraAt.apply(time));
```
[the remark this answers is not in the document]
thetaback: Fixed in [7b51da8](https://github.com/team401/coppercore/pull/213/commits/7b51da882e6fb79278c1e2cb6dcd9828bdadf9b2)

vision/src/main/java/coppercore/vision/VisionIOPhotonSim.java
```diff
@@ -65,14 +65,24 @@ public void updateInputs(
...
+        robotToCameraAt
+                .apply(Timer.getFPGATimestamp())
+                .ifPresentOrElse(
+                        (robotToCamera) -> {
+                            visionSim.addCamera(cameraSim, robotToCamera);
+                        },
+                        () -> {
+                            System.err.println(
+                                    "could not add camera as robotToCamera does not exist");
+                        });
```
aidnem: If this fails on code init (when it's running), which it likely will, as no data will be present in the maps, we just won't have simulated cameras, right? What if you just created them with a 0,0,0 transform3d and trusted them to get updated whenever the maps propagate?

aidnem: Fair points. We'll just have to make sure to add a value for the "default position" to the TimeInterpolatableBuffer in sim so that it has a value in it on code init, since this will run before any subsystem periodic has run and therefore no data will be present by natural means.

> Also, what do you mean by maps? Do you mean a possible TimeInterpolatableBuffer

Yes, I meant the TimeInterpolatableBuffer, sorry. In my head it's a "map of time to position" so that's what I called it.

PR #217, opened 2026-02-02
Add temperature status signal and separate low vs. high priority status signal refresh rates in MotorIOTalonFX

aidnem:
Verification:

- [ ] Publish to maven local on 2026 robot code and verify that this doesn't break anything

### DIFF
--- a/wpilib_interface/src/main/java/coppercore/wpilib_interface/subsystems/motors/talonfx/MotorIOTalonFXSim.java
+++ b/wpilib_interface/src/main/java/coppercore/wpilib_interface/subsystems/motors/talonfx/MotorIOTalonFXSim.java
@@ -14,8 +14,6 @@ import coppercore.wpilib_interface.subsystems.sim.CoppercoreSimAdapter;
 import edu.wpi.first.wpilibj.RobotController;
 import edu.wpi.first.wpilibj.Timer;
 
-// TODO: Add all signal group customizations to overloads in MotorIOTalonFXSim
-
 /**
  * A MotorIOTalonFXSim uses Phoenix-6 simulation features (TalonFX sim state) to read and update a
  * DCMotorSim, ElevatorSim, or SingleJointedArmSim periodically and propagate its values into the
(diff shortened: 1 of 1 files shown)


aidnem:
@PChild I should have clarified the javadoc a bit. The "low" priority group is all the signals that were at 50hz before but aren't high priority. All the signals we don't even mention in code will remain at 4hz, while the others will go from 50->20 and the high priority will go from 50->200/250.

aidnem:
Very low priority ticket here, but would be nice to merge eventually.

wpilib_interface/src/main/java/coppercore/wpilib_interface/subsystems/motors/talonfx/MotorIOTalonFX.java
```diff
@@ -184,10 +233,29 @@ protected MotorIOTalonFX(
                     closedLoopOutputSignal,
                     closedLoopReferenceSignal,
                     closedLoopReferenceSlopeSignal,
+                    temperatureSignal
                 };
+        final BaseStatusSignal[] highPrioritySignals = {
```
godmar: The obvious concern with this is that it hardwires which signals are "high" and which are "medium" priority.

I asked Claude to sketch what it would look like if this were programmable:

https://gist.github.com/godmar/a7f68da3ef54916a043e4dfc787958d8

(I have this in a worktree, if you like it I could commit and push it.)

godmar (changes requested):
At the least, explain what medium and high is.
Also explain why there's no low (the default was low?)

(PS: did I miss the explanation of 'low' in the javadocs?)

Possibly consider the option of making it programmable, see comment.

wpilib_interface/src/main/java/coppercore/wpilib_interface/subsystems/motors/talonfx/MotorIOTalonFX.java
aidnem: Oh, I think I missed how it's making an array big enough for all signals first and then copying into a smaller array later.

wpilib_interface/src/main/java/coppercore/wpilib_interface/subsystems/motors/talonfx/MotorIOTalonFX.java
```diff
@@ -45,6 +49,42 @@
  * MotionMagicExpo, MotionMagicVelocity, and TorqueCurrentFOC wherever possible.
  */
 public class MotorIOTalonFX extends CanBusMotorControllerBase implements MotorIO {
+    /**
+     * The default refresh rate for medium priority status signals such as applied voltage or closed
```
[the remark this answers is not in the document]
godmar: Done

godmar (approved):
LGTM

Issue #218, opened 2026-02-06
explore PUT functionality for constant http server


godmar:
We already have a `GET` to retrieve an object. Let's try adding a `PUT` to update the object; along with a `POST` to call a pre-registered function to apply updated values as necessary.

PChild:
Closed by #219

PR #219, opened 2026-02-06
(#218) Add HTTP PUT and POST support to JSONHandler

godmar:
- Add PUT method to update object fields from JSON request body
- Add POST method with preregistered callback that returns boolean
- POST response includes {"success": <bool>, "data": <object>}
- Add registerPostCallback() method with Function<T, Boolean> signature
- Add comprehensive unit tests including:
  - Basic PUT/POST functionality
  - Objects with WPILib unit fields (Distance, LinearVelocity)
  - Objects with arrays (List<Integer>, List<String>)
  - Objects with unit arrays (Angle[], Distance[])
- Bump version to 2026.2.8

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,2 +1,2 @@
-version=2026.2.7
+version=2026.2.8
 systemProp.file.encoding=utf-8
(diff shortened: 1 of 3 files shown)


PChild:
Can we get #217 merged before this goes in so we can have both in the 2026.2.8 release?

aidnem (approved):

Issue #230, opened 2026-02-15
make logging singleTagObservations optional


godmar:
The task of the vision code is to estimate poses from tags, ideally using multi-target pose estimations.  However, the current code also records observations for every single tag it sees, which is a function that was used in 2025 for aligning at the reef using a single tag.  (The function used for this was  `getDistanceErrorToTag`.)

Identifying and logging all SingleTagObservations is costly and shows up in VisualVM as having not insignificant time consumption; it should not be done unless requested.

Make it so that these SingleTagObservations are recorded only when requested.

Investigate whether the same applies to the array of fiducial ids.

(Another option is to remove the single tag code altogether with the idea that it would be recreated if needed again in the future).

PR #232, opened 2026-02-15
(#230) Added support for toggling single tag observation logging

thetaback:
- updated some of the java doc comments
- added an overload that can tell the camera whether or not its single tag observations should be logged


aidnem (approved):
Looks good to me. Do we need to bump version so we can publish?

Issue #233, opened 2026-02-17
Add default implementation for isLoggingSingleTags to VisionIO


aidnem:
When trying to upgrade to 2026.2.11 coppercore, in 2026 Robot Code, we get build failures because the base VisionIO (used in Replay) doesn't implement a default empty implementation for isLoggingSingleTags.

We can either:
a) Deal with it in robot code; this is maybe the right way because we can make the anonymous classes return the correct constant rather than a hardcoded value
b) create a default method implementation so that we only need to override it when we start logging single tags.

godmar:
I say do (a) as a temporary fix, but (b) the default implementation should return false since it's most certainly not logging single tags. @thetaback can you change that? We can pick it up on the next coppercore update.

PR #234, opened 2026-02-21
(#233) Added default implementation of isLoggingSingleTags

thetaback:


godmar (approved):
LGTM, let's schedule this with the next release.

PR #236, opened 2026-02-22
Make output status signals selectable via bitmask in MotorIOTalonFX

godmar:
Add SIGNAL_TORQUE_CURRENT and DEFAULT_OUTPUT_SIGNALS constants, and replace the hardcoded {motorVoltageSignal, torqueCurrentSignal} output array with signalsForMask(outputSignals), threaded through all constructors and factory methods. Default output group includes only torque current. Also fixes existing bug where public constructors accepting medium/high priority masks silently ignored them.


aidnem (approved):

aidnem (commented):
LGTM

Issue #246, opened 2026-02-26
Add arbitrary feedforward support to motor IOs


aidnem:
Based on initial testing of our shooter and turret, we'll likely need to use an arbitrary feedforward value for at least the shooter. We need to add methods to MotorIO to support the use of withFeedforward on our control requests with talonFX. This doesn't have to be supported by the spark IOs right now.

PR #247, opened 2026-02-26
(#246) Add arbitrary feedforward support

aidnem:
Adds support for providing an arbitrary feedforward to motor IOs. Implements this functionality for TalonFX only.

### DIFF
--- a/wpilib_interface/src/main/java/coppercore/wpilib_interface/subsystems/StatusSignalRefresher.java
+++ b/wpilib_interface/src/main/java/coppercore/wpilib_interface/subsystems/StatusSignalRefresher.java
@@ -39,10 +39,6 @@ public class StatusSignalRefresher {
      * CommandScheduler.getInstance().run()} in the robot periodic to work properly.
      */
     public static void refreshAll() {
-        if (canBusToSignalsMap.isEmpty()) {
-            return;
-        }
-
         for (var busInfo : buses) {
             // Refresh all signals on this bus and log the resulting status code
             var status = BaseStatusSignal.refreshAll(canBusToSignalsMap.get(busInfo.bus()));


aidnem:
> If so, perhaps 'additional' or 'supplementary' instead of 'arbitrary' might be a better choice?

Yes, according to the javadoc of the talonFX closed loop requests it's added in addition to the standard feedforward gain results. I've updated the javadoc to reflect this and renamed it to supplemental

godmar (approved):

godmar:
It looks like REV supports that as well, but is the difference that Spark's is specified in additional Volts whereas the TalonFX's is specified in additional Amps of current?

aidnem:
> The warning in the CTRE doc:
> 
> When setting and tuning closed-loop gains, Arbitrary Feed Forward should be set first, before any other values. The Arbitrary Feed Forward will change the relationship between your closed-loop gains and the output of your system, and thus result in different gains needed for a well-tuned mechanism.
> 
> makes me think it might not be.

This is true, but I'd still like to have it open as an option.

I can prototype option 3 if you like.

godmar:
To future proof it, option 3 might be best.

aidnem:
@godmar since we're now suddenly struggling with memory allocations, how would you feel about just having methods for setting FF voltage and current? Calling unsupported ones could print a warning and essentially be a no-op.

aidnem:
I had @PChild approve this since, as PR author, I physically can't click approve.

godmar:
> I had @PChild approve this since, as PR author, I physically can't click approve.

To me, in the UI, it appeared as "approved" so I thought you could just merge it, even after I snuck in [e3bde14](https://github.com/team401/coppercore/pull/247/commits/e3bde141ed6f8764756b1a60fa1a20e9634f68ca)

PR #250, opened 2026-03-03
(2nd PR): add statussignal refresher

aidnem:


aidnem:
Misunderstood where that extra commit went. Apologies. This PR is unnecessary.

PR #260, opened 2026-03-21
(#258) Add chirp to motorio

### LINKED ISSUE #258: Add chirp support to motor IOs
Since we likely won't have LEDs, we need some way to signal that we're ready to home. Add chirp support to play some audio from the motors once code has started up.

aidnem:


avidraccoon (approved):
Looks Good Enough

